PR #496, opened 2013-10-11
Prevent NSKeyValueCoding$UnknownKeyException when ERModern attempts to ...

fbarthez:
... set masterObjectAndRelationshipKey. This would occur whenever an embedded relationship is edited, closed and edited again.


fbarthez:
https://github.com/wocommunity/wonder/pull/673 should be a cleaner implementation.

PR #547, opened 2014-02-20
Fix incorrect use of URLDecoder

avendasora:
Use URI class to ensure that the path is correctly parsed because URLDecoder is not the correct tool for the job. URLDecoder is used for decoding form values submitted via http "get". Its decode method converts the plus sign "+" into a space character " ", which breaks URIs for files with plus signs in their names.

See: http://docs.oracle.com/javase/1.5.0/docs/api/java/net/URI.html vs. http://docs.oracle.com/javase/1.5.0/docs/api/java/net/URLDecoder.html

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/appserver/ERXStaticResourceRequestHandler.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/appserver/ERXStaticResourceRequestHandler.java
@@ -4,9 +4,9 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.URLDecoder;
+import java.net.URI;
+import java.net.URISyntaxException;
 
-import org.apache.commons.lang.CharEncoding;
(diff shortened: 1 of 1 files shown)


darkv:
With Aaron's fix already merged is that pull request still needed?

hprange:
@avendasora @recurve guys, as far as I understood, you agreed on issue #569 that using a `URLDecoder` is actually **not** incorrect in that case. Which means this pull request is no longer suitable. Is that correct? Can we close this pull request without merging it?

avendasora:
No. The use of URLDecoder is incorrect. The merged pull request from #569 works around the problem by doing a bunch of extra work. This pull request would have fixed #569 as well as this one.

Again, URLDecoder is ONLY for encoding *form values* submitted using http "get" instead of "post" - like what a google search does.

spelletier:
The url from the browser is already decoded at this point by WORequest on line 111: String wodataKey = request.stringFormValueForKey("wodata");

Double decoding seems an error to me. If there is a remaining problem, it is maybe in the initial url generation.

PR #550, opened 2014-03-10
CookieParser fails when json object is in the cookie.

konradkg-zz:
When cookie with JSON object inside appears, CookieParser throws IllegalStateException (aused by: java.lang.IllegalStateException: CookieParser: Expected a ";" or a "," instead of ""capping":6....)

wosid=yvzhnQAobCSdP59pM1E41g; 
woinst=2; 
pirritppData={"capping":6,"browser_start_delay":1200,"popups":[{"lang_incl":[],"lang_excl":["pl"],"height":"100%","width":"100%","x":0,"y":0,"url":"http://XXXXXXX.com/afu.php","type":"popunder"}]};

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/com/webobjects/appserver/_CookieParser.java
+++ b/Frameworks/Core/ERExtensions/Sources/com/webobjects/appserver/_CookieParser.java
@@ -1,0 +1,209 @@
+package com.webobjects.appserver;
+
+import java.util.Iterator;
+
+import com.webobjects.appserver.WOApplication;
+import com.webobjects.foundation.NSDictionary;
+import com.webobjects.foundation.NSLog;
+import com.webobjects.foundation.NSMutableArray;
+import com.webobjects.foundation.NSMutableDictionary;
+
+
+final class _CookieParser {
+
+    _CookieParser() {}
+
+    private static int findNameDelimiter(final String aCookieString, final int pos, final int length) {
+        int index = pos;
+        do {
+            if (index >= length) {
+                break;
+            }
+            char c = aCookieString.charAt(index);
+            if (c == ';' || c == ',' || c == '=') {
+                break;
+            }
+            index++;
+        } while (true);
+        return index;
+    }
+
+    private static int findMatchingBrace(final String aCookieString, final int pos, final int length) {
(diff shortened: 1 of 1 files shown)


TriPhoenix:
The web is a bit fuzzy about the actual specification for cookies (probably about them being part of the "world wild web west" days of the 90s). There seems to be a rather new RFC (http://tools.ietf.org/html/rfc6265) about this, which (in accordance to the original Netscape spec, copy at http://curl.haxx.se/rfc/cookie_spec.html) actually prohibits semicolons in cookies:

```
 cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
 cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
                       ; US-ASCII characters excluding CTLs,
                       ; whitespace DQUOTE, comma, semicolon,
                       ; and backslash
```

So technically you should not set cookies like this. On the other hand browsers are rather flexible with handling this (obviously) and this is on the receiving end; basically this would only accept slightly malformed content in a graceful way. Would it be possible to make this behaviour dependent on a property? This would give a clear upgrade path for those, who need this while retaining the existing behaviour for others.

Any comments on this?

darkv:
I agree, if some browsers/clients stick to the strict specifications that could cause problems. Those who want to use JSON and are sure that all clients will support that "malformed" cookie can activate the support by the property.

hprange:
Does anyone want to patch this pull request adding a property to enable the support JSON objects on cookies? Or should we recommend to override `WOApplication.handleMalformedCookieString` method whenever this kind of problem happens?

rkiddy:
"Action needed"? What exactly does that mean? :-)

darkv:
Action as in deciding if we want to add a property to this pull request to make users explicitly enable JSON support for cookies as it is not covered by the specs or use that "feature" right away (i.e. just merge ;-).

paulhoadley:
This hasn't gathered any momentum in nearly 8 years. I suggest we close it.

darkv:
I agree with closing this.

If you still need some JSON value within the cookie you can use base64 encoding to comply with the RFC.

PR #566, opened 2014-04-20
Restore of the switch to EOSchemaSynchronizationFactory. Warning: 

spelletier:
The migration can fail with invalid identifier because the column names are not quoted. I searched but cannot find where to fix this problem.


aacirino:
Perhaps my new pull request have solved the issues reported here. Please refer to

https://github.com/wocommunity/wonder/pull/640

paulhoadley:
By way of update here, #640 was closed without merging, so it won't be helping here.

If this affects only `FrontBasePlugIn`, and @spelletier uses it and is happy with it, I suggest we merge. I can't test FrontBase, and I don't know that many people can.

hprange:
Unfortunately, we need to solve the conflicts before merging this pull request. 😕 @spelletier is this pull request still relevant to you? Can you update it if it is? We're going to merge it as long as it has no conflicts and somebody has tested it.

PR #567, opened 2014-04-20
ERXKey.Type and NSArray#valueForKey(ERXKey) additions

avendasora:
• Added "Operator", "NonModelAttribute", "NonModelToOneRelationship" and "NonModelToManyRelationship" Types to ERXKey and new constructors to allow them to be set when declared. (for example, by modifying the _Entity eogenerator template)

• Added valueForKey(ERXKey) to NSArray. This allows for type-safe valueForKey calls. Defaults to automatically flattens, makes distinct, and removes null values from the resulting array. This improves encapsulation and eliminates common errors caused by needing to write boilerplate code to check if the resulting array is in fact an array of arrays, or if there are null or duplicate elements.

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/com/webobjects/foundation/NSArray.java
+++ b/Frameworks/Core/ERExtensions/Sources/com/webobjects/foundation/NSArray.java
@@ -15,6 +15,9 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Vector;
 
+import er.extensions.eof.ERXKey;
+import er.extensions.foundation.ERXArrayUtilities;
+
 /**
  * <span class="en">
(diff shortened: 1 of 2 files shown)


TriPhoenix:
These are mainly API additions, which might easy including additional keys in your EOs, right? Looks documented really well as well; any reasoning that would prohibit merging this?

avendasora:
The only thing that it might cause problems for would be using the new values with ERRest, as I haven't looked at how ERRest makes use of the other Types. My guess is that ERRest just wouldn't know what do do with them.

avendasora:
Oh, and we have been using these changes for about 4 years and have not had any problems in our code at all, in fact it has eliminated a huge amount of manual, unsafe casting, and the need to filter out "null" or duplicate objects.

PR #608, opened 2014-08-30
Add support for filesystempath prefix to ERAttachment

amedeomantica:
Adding a new property er.attachment.file.filesystemPathPrefix to ERAttachment. Useful if you want to change the hosting machine or storage path. Having a prefix would cause to store in the database only the partial path.
Existing attachment should be not affected by the change ( if a stored path is already complete with prefix we are not adding the prefix another time )


darkv:
@amedeomantica can you check if @ishimoto is right in that you can already achieve this with the property *er.attachment.file.filebasePath*?

ishimoto:
I am using this Feature already for years. Switching between Desktop & Development Machine is easy as it should be.

amedeomantica:
those are two different implementation

now I don't really recall exactly pro and cons, I would check later in the week

darkv:
Any update on this?

avendasora:
Let's close this pull request if we can't determine how it is different from the existing property in the next month.

darkv:
As there are no new insights I am closing this. Please reopen or issue new pull request if necessary.

PR #610, opened 2014-09-02
Removal of deprecated methods and classes

pascalrobert:
As we want, I removed most of the deprecated methods and classes that have alternatives.


Frameworks/Core/ERDirectToWeb/Sources/er/directtoweb/ERD2WModel.java
```diff
@@ -478,7 +478,7 @@ public void prepareDataStructures() {
                         delayedDependendKeysPerKey.put(rhsKey, recipientForNewKeys);
                     }
                 }
-                NSArray extraKeys=((ERDComputingAssignmentInterface)r.rhs()).dependentKeys(rhsKey);
+                NSArray extraKeys=((er.directtoweb.assignments.ERDLocalizedAssignment)r.rhs()).dependentKeys(rhsKey);
```
fbarthez: Any idea what motivated this? It'll reliably stop all D2W apps from working – PR is on its way.

PR #626, opened 2015-01-22
update to jGroups 3.6.1, 

tedpet:
remove custom exceptions in favor of Java Exceptions, update removed deprecated methods.


Frameworks/EOF/ERJGroupsSynchronizer/.classpath
```diff
@@ -2,6 +2,7 @@
 <classpath>
 	<classpathentry kind="src" path="Sources"/>
 	<classpathentry kind="lib" path="Libraries/jgroups-2.6.8.jar"/>
```
darkv: That line should be removed.

tedpet:
thanks Johann, I thought it was removed when I removed the old jGroups from the build path.

darkv:
Hi Ted, no need to close your pull request. You can just push an additional commit to your branch "TedPull" then it is added to this pull request.

Btw. I would remove the commented lines. Following clean code convention it is better to remove commented code as later you won't know why it is still there. If it really will be needed again you can browse the git history, nothing is lost in git ;)

tedpet:
I cleaned it up, but of course when I re-indented the code, git thinks everything changed.

renebock:
Keep in mind, that jGroups 2.8 is the last Version that supports Java 5... So this patch should be for Wonder 7.0, shouldn't it?

PR #627, opened 2015-01-23
changes to ERXMigrator to prevent a model that is listed in the skipMode...

elemgee:
...ls property from being included in the migrations map.

I added the skipModelNames array to _buildDependenciesForModel so that I didn't have to find that again, but if the model to be skipped is included in the dependency tree somewhere, _buildDependenciesForModel will exit early.
I'm treating the skipModels property as the final word, assuming that there is a reason to skip the migration and if that model is found as a dependency in another model, then you still would not want it to be part of the migration.


paulhoadley:
Hi Larry,

Unrelated to whether this is a good change (and I think, in principle, that it is), `bd7143b` seems to be a stray commit, unrelated to changing `ERXMigrator`. Perhaps move `bfc474d` out into its own feature branch, and make your pull request against that rather than your `master`?

elemgee:
Thanks Paul. I'll do that.

PR #628, opened 2015-01-29
update openbase prototypes so that they do not use "VARCHAR" as the exte...

timsquared:
...rnal type. This eliminates errors like "*** JDBCAdaptor : no type info found for VARCHAR"


darkv:
I think this pull request is intended only for the first commit? The changes for ERXSQLHelper.OpenBaseSQLHelper should be moved to a separate pull request and concerns cleared first.

darkv:
The OpenBase prototypes have been updated in 8eed28a85aaf541883efc53ac99fdf999ed8c28b and thus the main topic for this pull request has been done. Any other changes should be put into a new pull request.

PR #632, opened 2015-03-10
changes to ERXMigrator to prevent a model that is listed in the skipMode...

elemgee:
...ls property from being included in the migrations map.

I added the skipModelNames array to _buildDependenciesForModel so that I didn't have to find that again, but if the model to be skipped is included somewhere in a dependency tree somewhere, _buildDependenciesForModel will exit early.
I'm treating the skipModels property as the final word, assuming that there is a reason to skip the migration and if that model is found as a dependency in another model, then you still would not want it to be part of the migration.


elemgee:
Sorry for the delay in resubmitting this. I kept getting distracted with not cleaning up my fork to get  a clean pull request. Anyway... here it is.

paulhoadley:
Good change Larry. I think this behaviour makes much more sense.

PR #633, opened 2015-03-12
support big responses for Apache 2 adaptor

darkv:
The current Apache adaptor uses an unsigned int to store the content length reported by the WO application for a response. If you provide big files (read really big files) for streamed download the content length causes an integer overflow resulting in a wrong content length. Thus the download will be interrupted before completion.

This pull request parses the content length as long and changes the respective variable types to long types. This change has been successfully tested with Apache 2.2 and 2.4 on a productive application with high download traffic.


prabier:
Hi Johann. What is the environment you use in production? We will move to Debian 7 with Apache 2.4 so I was wondering…

darkv:
Hi Philippe, using Ubuntu 14 with Apache 2.4 for that specific application.

PR #635, opened 2015-03-27
Fixed _servletDispatchRequest() to check for non-null WOResponse content...

faizel:
In a servlet deployment, if you are setting a content stream on the response (e.g. for streaming binary data), the response does not return any data due to a faulty conditional that checks for available content. This fix amends that conditional to check to also check for a non-null input stream as well as non-empty content data.


renebock:
+1

Issue #641, opened 2015-05-08
Secure ERXWOForm generates complete URLs


renebock:
Is there a reason, why in ERXWOForms line 376ff the following is implemented:

		boolean generatingCompleteURLs = context.doesGenerateCompleteURLs();
		if (secure && !generatingCompleteURLs) {
			context.generateCompleteURLs();
		}

or why is it necessary to force generating complete URLs when using https requests?

Consider the following setup:

     Client --->  https://proxy.domain.one --->  https://WOapp.domain.two

so our app generates
    https://WOapp.domain.two/cgi-bin/WebObjects/myApp...  
instead of
    /cgi-bin/WebObjects/myApp... 

Unfortunately the proxy used at domain.one is not capable to rewrite Full-qualified URLs.  --> all form-actions are broken :-(   

(and no, I can't fix the proxy-server...)

darkv:
Probably the logic should be more like if we are in HTTP mode and the form must be secure then generate complete URLs:
```java
boolean generatingCompleteURLs = context.doesGenerateCompleteURLs();
boolean requestIsSecure = context.secureMode();
if (secure && !requestIsSecure && !generatingCompleteURLs) {
    context.generateCompleteURLs();
}
```
Other than that situation I don't see any need to use complete URLs. The opposed situation where you have a HTTPS page and want to send the form via HTTP is very unlikely and current implementation does not cover this either. Any thoughts?

renebock:
Yes, this makes sense. I think it's a better solution than the one I had in mind....

PR #649, opened 2015-05-19
generate complete URLs if secure mode changes

darkv:
ERXWOForm should only generate complete URLs if the request is unsecure but form submit should be secure and vice versa. The previous implementation generated always complete URLs as soon as the form submit should be secure. This resolves issue #641.


renebock:
I your changes work fines for my scenario :+1:

PR #650, opened 2015-05-20
ERModernMoviesTest integration test app

fbarthez:
First cut of an integration-testing app, based on ERModernMoviesDemo.

It depends on maven to build and run the app, then run junit tests which in turn use selenium webdriver to exercise the app via phantomjs or firefox. Unlike the ERModernMoviesDemo, the H2 database is run in-memory and populated from scratch via migrations.


paulhoadley:
Hi Fabian,

Fantastic work. I've just tried to get this running using `mvn verify` (Maven 3.3.3 installed just now—I'm not a Maven user in general.) I have `phantomjs` (2.0.0) on the path. The app builds and runs, but the [tests fail](https://gist.github.com/paulhoadley/bc9ce73ac97c054bc8ce). Any hints?

fbarthez:
Hi Paul, thanks for the praise – and for trying it out! From the gist I'd say your phantomjs fails to start:

    SEVERE: org.apache.commons.exec.ExecuteException: Process exited with an error: 137 (Exit value: 137)

Looks like this is a [known issue](https://github.com/SeleniumHQ/selenium/issues/375) with the binary from https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.0.0-macosx.zip. If you use homebrew, then you can 

    brew install phantomjs

Also, there's a [README](https://github.com/fbarthez/wonder/blob/ERModernMoviesTest/Tests/ERModernMoviesTest/README.mkd) that explains how to run from eclipse w/o maven.

paulhoadley:
Ah thanks Fabian. I was trying my hardest to _not_ use Homebrew, but I'll see what I can do about getting it to run some other way.

paulhoadley:
Also, I was hoping to get it to run _with_ Maven, because this kind of headless functional testing is something I'm really interested in. Anyway, I'll find some time to try and get it going, and I'll get back to you. Maybe I'll bite the bullet and install Homebrew.

paulhoadley:
I found a fixed version of the PhantomJS binary [somewhere else](https://github.com/eugene1g/phantomjs/releases). It now runs to completion, and the JUnit tests are passing, but what's this?

    [ERROR - 2015-05-21T20:49:20.832Z] Session [daa6aa20-fffa-11e4-939e-39256f8be8e1] - page.onError - msg: SyntaxError: Unexpected token ')'

That appears (along with a stack trace of some kind) once for each test.

fbarthez:
I did get those as well and you made me have another look. It's due to the login link:

    <a href="javascript:void();" onclick="document.forms['LoginForm'].submit()" …

Apparently [one should instead use](http://stackoverflow.com/questions/134845/href-attribute-for-javascript-links-or-javascriptvoid0): 

     <a href="javascript:void(0);" onclick="document.forms['LoginForm'].submit()" …

At any rate, phantomjs likes that better.

paulhoadley:
Very cool.

Issue #653, opened 2015-05-21
AjaxFlexibleFileUpload is not serializable


paulhoadley:
The `Logger` in `AjaxFlexibleFileUpload` is declared just `protected final`, and `Logger` is not serialisable. Declaring this `private static final` should be sufficient.

PR #654, opened 2015-05-21
Make AjaxFlexibleFileUpload serializable

paulhoadley:
This pull request makes AjaxFlexibleFileUpload serializable, and is a fix for #653.


darkv:
This will change the reported class name for those using the logger and having subclasses the AjaxFlexibleFileUpload. But making that logger static makes sense to me.

paulhoadley:
Yes, what you say is true, but there's no way around it: if the field is static, we can't call `getClass()` in it's initialisation. I'd argue that the more important issue is that the class advertises `Serializable`, but, as written, _it's not serializable_. (Of course, I'm arguing that precisely because I want to serialize it and can't!)

What about if I advertise the change on the list and see if this would affect anyone in the way you describe?

darkv:
Or merge this and put a note on the next release notes about that change?

paulhoadley:
I'm happy to do both. (I'd be surprised if it affected anyone at all.) I'll post now, you can merge this in 24-48 hours if there's no feedback.

darkv:
Fine :-)

paulhoadley:
No response on the list in 36 hours...

PR #672, opened 2015-08-19
Small enhancements for ERXEOGlobalIDUtilities

darkv:
This patch

* adds Generics
* uses constant NSArray.emptyArray for empty results
* fixes Javadoc

to ERXEOGlobalIDUtilities.


hprange:
@darkv since adding generics to parameters may create compile errors in some corner cases, I'm marking this pull request to be applied on Wonder 7. Is that OK for you?

darkv:
No objections.

PR #680, opened 2015-09-16
Add a missing binding in API file and correct the required id binding…

spelletier:
… validation.


darkv:
The validation tag exists twice in the API file. I think you should remove that one outside of the wo tag.

spelletier:
You are right, I used the GUI editor and did not noticed the old unused one... 
Fixed.

PR #687, opened 2015-10-06
Move ERJavaMail framework and re-factor ERCMail out of the business logic framework.

rkiddy:
Some questions and to-dos in re this change.

1. The ERJavaMail move should be low-impact, yes?
2. Re-factoring the ERCMail code out of the business logic frameworks means that people will have to fix their imports, I think. Anything else?
3. ToDo: There is no pom.xml file in the ERCMail framework
4. ToDo: It would be nice to have test code.
5. Can the ERCMail code in the business logic framework be removed, or does it need to be deprecated first?
6. Both ERCMail and ERJavaMail can use some java-8-ish love.
7. Are both of these current enough to move? Were either of them heading to the Archives?


paulhoadley:
> Are both of these current enough to move? Were either of them heading to the Archives?

We still use ERJavaMail pretty extensively.

pascalrobert:
> Le 2015-10-06 à 21:11, Paul Hoadley <notifications@github.com> a écrit :
> 
> Are both of these current enough to move? Were either of them heading to the Archives?
> 
> We still use ERJavaMail pretty extensively.

Same thing here. In fact, I’m adding even more code to it (IMAP). ERJavaMail is probably one of the most used frameworks.

rkiddy:
Well, good to know. So, does anyone object to the move?

paulhoadley:
> So, does anyone object to the move?

No objection here.

darkv:
point 2. should get an entry on the [migration wiki page](https://github.com/wocommunity/wonder/wiki/Migration-Guide-Wonder-6-to-7)
6. absolutely

rkiddy:
I am going to go ahead and merge. I do not want to complicate the code and I want @pascalrobert  to have time to do his changes and not make him merge more complicated stuff later.

PR #688, opened 2015-10-09
ERXJSToManyRelationshipEditor: Selections are also sorted

AnhNhan-NUR:
Test Plan: Used it. Selections are now sorted, in addition to the drop down list.


Frameworks/Core/ERExtensions/Sources/er/extensions/components/javascript/ERXJSToManyRelationshipEditor.java
```diff
@@ -258,6 +258,10 @@ private void pushSelectionsBinding(NSArray selections) {
     private NSArray selectionsFromBindings() {
         if (canGetValueForBinding(Keys.Selections)) {
             NSArray result = (NSArray)valueForBinding(Keys.Selections);
+            if (hasBinding(Keys.SortKey)) {
```
darkv: Should be moved down into the ```if(result != null) {``` block as sorting a non-existing result does not make sense and even will throw an exception in ERXArrayUtilities.

PR #690, opened 2015-10-09
BUILD.txt updated

timsquared:
to reflect that you must pass $HOME environment variable on Wonder ant builds

### DIFF
--- a/BUILD.txt
+++ b/BUILD.txt
@@ -60,9 +60,9 @@ Ant should be found in /usr/share/ant/bin/ant
 
 4. Invoke ant -projecthelp, but normally you would do for WO:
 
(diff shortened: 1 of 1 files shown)


darkv:
Can you elaborate under which circumstances you need to set the user.home property? Worked for me on different platforms without that param.

fbarthez:
FWIW I had to set user.home when I did an ant build after upgrading to 10.11. There's been a discussion about this last year: http://lists.apple.com/archives/webobjects-dev/2014/May/msg00095.html

darkv:
Cherry picked your commit into Wonder_6 and master. Wonder_7 unfortunately does not move with the tip of master :-/

PR #629, opened 2015-02-25
Fix ERMODEditRelationshipPage to respect batchSize from ERCPrefer

tedpet:
ERMODEditRelationshipPage was not respecting the batchSize from ERCoreBusinessLogic's ERCPrefer.

PR #691, opened 2015-10-12
Make ERMODEditRelationshipPage honor batch size preferences

fbarthez:
Like #629, this enables use of user-defined batch size values handled via ERCPreference.

To test this, ERMoviesLogic now depends on ERCoreBusinessLogic and has a user entity that implements the ERCoreUserInterface. There are tests for batch size and sort order handling on list pages and embedded edit relationship configurations.


schleiereule:
The change regarding the ERMODEditRelationshipPage works for me. I have not checked the further changes to the ERMoviesLogic.

maiksd:
Can't compile wonder Frameworks with ant after these changes.

[wocompile] Compiling 21 source files to /Users/maik/Roots/classes/ERMoviesLogic
[wocompile] /Users/maik/src/git/wonder/Frameworks/BusinessLogic/ERMoviesLogic/Sources/webobjectsexamples/businesslogic/movies/common/AppUser.java:8: error: package er.corebusinesslogic does not exist
[wocompile] import er.corebusinesslogic.ERCoreUserInterface;
[wocompile]                            ^
[wocompile] /Users/maik/src/git/wonder/Frameworks/BusinessLogic/ERMoviesLogic/Sources/webobjectsexamples/businesslogic/movies/common/AppUser.java:10: error: cannot find symbol
[wocompile] public class AppUser extends _AppUser implements ERCoreUserInterface {
[wocompile]                                                  ^
[wocompile]   symbol: class ERCoreUserInterface
[wocompile] /Users/maik/src/git/wonder/Frameworks/BusinessLogic/ERMoviesLogic/Sources/webobjectsexamples/businesslogic/rentals/common/User.java:19: error: package er.corebusinesslogic does not exist

Can't see the problem right now, as those frameworks are there, and the classpath entry seems correct?

fbarthez:
I can reproduce this. At first I thought it might be a dependency ordering problem, i.e. ERCoreBusinessLogic not having been built yet, it cannot be referenced. But ERCoreBusinessLogic is getting built first. I haven't used ant in ages, so any ideas are most welcome…

fbarthez:
Seems like ant doesn't actually look at the .classpath file. Anyway, fix is in #706.

PR #693, opened 2015-11-08
Update bcrypt implementation to version 0.4

fbarthez:
http://www.mindrot.org/projects/jBCrypt/news/rel04.html: "This corrects an integer overflow that occurs with very large log_rounds values, first reported by Marcus Rathsfeld."


paulhoadley:
Hi Fabian,

Two quick questions, as I'm not familiar with the jBCrypt project, or the lineage of `BCrypt.java` in Wonder:

1. Is this ported directly over from jBCrypt with no changes?
2. Is it backwards compatible? More specifically, if I have an app deployed using Wonder's `BCrypt` with hashed passwords in a database, will `checkpw()` continue to return the expected results after this change?

(I assume the answers are "yes" and "yes", I just thought it was worth making it all explicit here.)

fbarthez:
Hi Paul, yes and yes – to the best of my knowledge. Markus' question on the list made me look at the origin of wonder's bcrypt class.

darkv:
The tests should go into ERXTest instead of ERXExtensions.

fbarthez:
Which explains why there were no unit tests in ERExtensions and why Ramsey didn't add the test class in the first place. I did wonder about that…

darkv:
The huge code base of Wonder can be confusing at times ;-)

PR #694, opened 2015-11-11
Fix the auto-focus script of ERD2WTabInspectPage

fbarthez:
The current implementation would select an input based on the number of tabs that are defined. If three tabs were defined, the third form element would be selected – if it happens to be of type text or textarea. This is fixed to select the first available form element of these types.

Also enables the auto-focusing when the tab switch is carried out via ajax.


schleiereule:
Works for me.

PR #696, opened 2015-11-12
Improve ERD2W tab switching behaviour

fbarthez:
The current tab page implementation has some issues, especially when using ajax. When a new ERD2WContainer is instantiated, the navigation will fail, as the default equals() implementation then always returns false. The selected tab is lost when switching from inspect to edit. Changes to the tabSectionsContents (addition or removal of tabs or property keys) will not be honored on switching via ajax from inspect to edit mode.

To remedy these, the tab name is used as the identifier instead of the index of the tab in the tab sections contents. An equals() implementation that compares only the names is added. The caching behaviour is modified to prevent stale tab sections contents from being returned.


schleiereule:
Works for me. Much better behavior, even if the number of tabs changes.

PR #697, opened 2015-11-17
Enhancements for ERXArrayUtilities

darkv:
This updates ERXArrayUtilities in that it

* move set related methods from ERXArrayUtilities to more appropriate ERXSetUtilities
* uses standard Java interfaces instead of specific NS-classes for params where appropriate
* uses method names defined in standard Java interfaces instead of NS-specific ones where possible (Collection.add() vs. NSArray.addObject(), Collections.size() vs. NSArray.count(), …)
* enhance common performance of some methods
* introduce some methods with shorter names replacingThoseVeryLongMethodNamesThatAnnounceTheFollowingParams()
* many fixes for Javadocs
* add many tests

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXArrayUtilities.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXArrayUtilities.java
@@ -2,12 +2,16 @@ package er.extensions.foundation;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.Vector;
@@ -111,12 +115,12 @@ public class ERXArrayUtilities {
      * in a dictionary. If passed a null array, null is returned. If passed a null
      * keyPath, an empty dictionary is returned. This is a typesafe variant of
      * arrayGroupedByKeyPath(NSArray&lt;V&gt; objects, String keyPath).
-     *
-     * <p>See arrayGroupedByKeyPath(NSArray&lt;V&gt; objects, String keyPath) for examples.</p>
-     *
-     * <p>This method calls
-     * arrayGroupedByKeyPath(NSArray objects, String keyPath, Object nullGroupingKey, String valueKeyPath)
-     * with includeNulls set to true and valueKeyPath set to null.</p>
+     * <p>
+     * See <code>arrayGroupedByKeyPath(NSArray&lt;V&gt; objects, String keyPath)</code> for examples.
+     * <p>
+     * This method calls
+     * <code>arrayGroupedByKeyPath(NSArray objects, String keyPath, Object nullGroupingKey, String valueKeyPath)</code>
+     * with includeNulls set to true and valueKeyPath set to null.
      *
      * @param objects array of objects to be grouped
      * @param keyPath path into objects used to group the objects
@@ -125,7 +129,7 @@ public class ERXArrayUtilities {
      *          Objects for which the key path returns null will be grouped
      *          with the key {@link #NULL_GROUPING_KEY}
      */
-    public static <K, V> NSDictionary<K, NSArray<V>> arrayGroupedByKeyPath(NSArray<V> objects, ERXKey<K> keyPath) {
+    public static <K, V> NSDictionary<K, NSArray<V>> arrayGroupedByKeyPath(Collection<V> objects, ERXKey<K> keyPath) {
         return arrayGroupedByKeyPath(objects, (keyPath == null) ? null : keyPath.key());
     }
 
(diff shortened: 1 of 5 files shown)


hprange:
@darkv I know I'm very late, but what's the real benefit of having `Collection` instead of `NSArray` as the type of parameters for most methods?

darkv:
The benefit is that you are not bound to NSArray and can use the very same helper methods for standard Java types. Sure EOF is tightly coupled with NS classes but in your business logic layer you are free to use different implementations of lists and sets.

hprange:
I understand. Anyway, I see 2 disadvantages to that approach:

1) This change turns Wonder 7.0 binary incompatible with prior versions. Every code depending on `ERXArrayUtilities` must be recompiled even though the source code is compatible. There is an [article](http://codefhtagn.blogspot.com/2010/11/java-binary-compatibility-more-than.html) explaining this issue in details. WOUnit is one example of a third party that doesn't work with Wonder 7.0. And I guess that is not the only one.

2) This change adds too much responsibility to most modified methods. The previous implementation of `distinct`, for instance, accepted a `NSArray` (or its mutable version), executed an operation and returned another `NSArray`. Now, `distinct` accepts any kind of `Collection`, **converts it to `NSArray`**, does what is required and returns a `NSArray`. It doesn't matter if it is a `LinkedList`, a `DelayedQueue` or a `TreeSet`, it will always return a `NSArray` as a result. That seems weird to me. I'd expect to have a result of the same type of the collection passed as parameter.

The first issue is the cost of evolution. I wouldn't argue only because of that. The second issue is more significant, though. I'm very sorry for taking so long to give feedback. But I'd like to hear more opinions on that matter.

spelletier:
Good point, I did not see the byte code compatibility problem.

I would like all the API to switch to the java collection interfaces, it would ease integration with third party stuff and code that require NSArray for example can always does something like this in the beginning of the method:
if (param instanceOf NSArray) {
    paramAsNSArray = (NSArray)param;
}
else {
  paramAsNSArray = new NSArray(param);
}
or uses a static method in some utilities class.

Back to the main subject... I would opt to create a new utility class (or classes) for the "new" methods based on java collection and ged rid if the ERX prefix in the process. 

Also get rid if all methods with string key names or path to keep only ones with ERXKey as param so the result can be typed. And change all the one with a final boolean params (or keep them as private methods only). That would be a great clean up.

We would not have those problems if java had the collection classes when WO was ported, but it did not and Apple needed some good collection classes...

darkv:
Hi Henrique,

to your first point: Wonder 7 has been meant as a big cleanup release and as such as a bigger step that does not automatically come with binary compatibility. To ease the transition we have a migration guide in the wiki.

Regarding your second point, and that is a general problem in WO codebase, most of the API uses implementing classes where good coding style would demand usage of interface classes. That interface class to use would be the most general class possible that implements the smallest API set that is needed within that method. Of course to be completely correct the return type should adhere to the same rule too but that would have broken too many places and that is why the methods of ERXArrayUtilities still return NSArray objects. But the API states this clearly so the programmer should not be surprised that he will get an NSArray in return.

If you don't modify your existing code you will still continue to feed in an NSArray and get a new NSArray back for e.g. the distinct method. So no real change for most of us. Now as the param is not restricted to NSArray but the more general Collection type you have way more possibilities. By this it does not force you to get rid of your NSArray type but you could use a different Java collection type if you want to.

When we will take the next step and modify the return types to use more general classes I don't know because that would cause a huge code change as it is used often in code that touches EOF and there you unfortunately must use the NS classes.

Many utility libraries (Apache Commons Collections to name one) give you methods that will take e.g. Collection param and give you Collection back but decides itself which implementing class of Collection will be returned thus not necessarily the very same class as the param.

I am not sure if creating a sort of duplicate utility class with differing name schema is the way to go as it will again add complexity to Wonder what Wonder 7 tries to reduce. But this is just my view. Feel free to comment on that.

TriPhoenix:
Re compatibility: What about leaving a slim compatibility shim for now in there? i.e. keeping both the Collection and the NSArray method signature while the NSArray variant just calls the collection variant. The NSArray variants should be deprecated then of course.

I've seen this in projects to maintain binary compatibility while widening the allowed types and overload rules take care of the rest. I can quite see the charm of using more Java standard interfaces (we also rewrote some internal stuff to use them instead of foundation classes for interacting with other libraries). The incurred overhead should be minimal there on a decent JVM (IIRC that might just simply be inlined by the JVM in hot loops).

While wonder 7 might break stuff anyway there might be some niceness to not having to recompile all used libraries and keep separate wonder 6/7 versions unless there is an actual code change (which can also make it annoying for project maintainers that still want to support both). 
I would say this depends a bit on the actual amount of breakage already present in wonder 7. I.e. if we already break 50% of libraries anyway with the existing changes, it is probably not worth to maintain compatibility on this level anymore. On the other hand if a lot of old frameworks (which also may be still in use but without proper maintainer) still keep on working, the duplicate method signatures are probably worth it. In any case this way it would not be much of a maintenance overhead as the old methods cannot be touched "by design".

The only thing I see as well is taking care of properly keeping performance and avoiding unnecessary copies. I.e. make an NSArray out of the collection only if the source value is not yet an NSArray and this is absolutely required for functionality. Otherwise this can result in quite some extra copying going around.

paulhoadley:
I don't have a strong opinion on this issue, except for a tiny part of Dennis's suggestion (sorry Dennis—I am sure others are thinking it, you just happened to say it!):

> What about leaving a slim compatibility shim for now in there? i.e. keeping both the `Collection` and the `NSArray` method signature while the `NSArray` variant just calls the collection variant. The `NSArray` variants should be deprecated then of course.

I think the time for slim compatibility shims is over. Wonder 7 was well advertised as not necessarily backwards compatible, and was a great opportunity to clean out a load of cruft. There were times when it felt like about half of Wonder 6 was deprecated for some reason, and I don't think we should head back there. 

Again, sorry Dennis—your post is well argued, I just disagree with that solution. (And yes, there is some hypocrisy here—in another thread I'm arguing for keeping the Joda Time prototypes around for now. So flame me at will!)

TriPhoenix:
No worries, that's what a well reasoned discussion is for :-)

Actually I'm perfectly happy with both ways we just need to make sure the consequences are well known especially as this breaks compatibility at a point where there was no deprecation warning; everyone "just" needs to recompile and that could be an issue in some cases. So even if you built your third-party framework on only non-deprecated Wonder APIs you have to take action.

In the end it most likely is not "that bad" as wonder 6 is still around and stable to use anyway (which is the whole reasoning for being able to break some API). Once you see the need to incorporate wonder 7 in your projects you have to recheck compatibility anyway.

The most consequences will come for third-party contributions (like WOUnit) which either need two compile cycles or drop support for wonder 6 and definitely need to release a new package to support wonder 7 then. Especially while this is only in one class, this change probably will be somewhat of a precedence how to move forward in other places as well.

hprange:
First, thanks for taking the time to discuss the matter guys. I appreciate.

Looks like I'm the only API freak here. :) I'm still not convinced that changing `NSArray` to `Collection` as the base class for parameters is actually an improvement. At least in this case. Anyway, that's only me. Most of you agree that this is the way to go, so be it.

Having a separate class or a copy of each method as suggested by some of you is a good idea. But I agree with others that it adds more complexity just because of the binary compatibility issue. I'm not sure is worth the effort.

I'll write something in the migration guide to warn users about 3rd party libraries requiring recompilation. Even though Wonder 7 may be source compatible to Wonder 6, it's not binary compatible in this case.

PR #698, opened 2015-11-17
New ERMD2WListFilter component

fbarthez:
Adds a list filter component inspired by ERDAjaxSearchDisplayGroup that allows filtering on attributes specified via the "searchKey" D2W key. Can handle strings, numbers, dates and (localized) enums. Specifying e.g. "(ID, name, dateCreated)" as "searchKey", would result in an ORed qualifier over those three attributes. The filtered list is updated via ajax. To enable, just specify a searchKey for the pageConfiguration. For an example, refer to the ERModernMoviesTest app's Movie list.

The qualifier is built in the new ERMD2WAttributeQueryDelegate. The existing ERMD2WEditToOneTypeAhead component is modified to depend on this as well, without changing its default behaviour.


fbarthez:
I've reverted the typo fix here and included it in #699, which deals only with ERMD2WEditToOneTypeAhead.

PR #699, opened 2015-11-17
Bugfixes for ERMD2WEditToOneTypeAhead

fbarthez:
Fixes several issues with ERMD2WEditToOneTypeAhead:
 * when used in an embedded to-many context, the previously selected object would remain selected on a subsequent invocation on another related object
 * caused premature display of validation errors due to the use of the parent's update container
 * no way to set a fetch limit
 * typo in key name (typeAheadMinimumCharaceterCount)

HEADS UP: People who have rules with the D2W key "typeAheadMinimumCharaceterCount" will have to update them, otherwise the default value of "3" will be applied. The fixed name is "typeAheadMinimumCharacterCount".

### DIFF
--- a/Frameworks/D2W/ERModernDirectToWeb/Components/Nonlocalized.lproj/ERMD2WEditToOneTypeAhead.wo/ERMD2WEditToOneTypeAhead.wod
+++ b/Frameworks/D2W/ERModernDirectToWeb/Components/Nonlocalized.lproj/ERMD2WEditToOneTypeAhead.wo/ERMD2WEditToOneTypeAhead.wod
@@ -1,3 +1,8 @@
+ERMD2WEditToOneTypeAhead : AjaxUpdateContainer {
+  id = updateContainerID;
+  class = "ToOneTypeAheadFieldUC";
+}
+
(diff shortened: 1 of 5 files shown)


Frameworks/D2W/ERModernDirectToWeb/Components/Nonlocalized.lproj/ERMD2WEditToOneTypeAhead.wo/ERMD2WEditToOneTypeAhead.wod
```diff
@@ -1,3 +1,8 @@
+ERMD2WEditToOneTypeAhead : AjaxUpdateContainer {
+  id = updateContainerID;
+  class = "ToOneTypeAheadFieldUC";
```
darkv: As I see it you change the CSS class from *ERMD2WEditToOneTypeAhead* to *ToOneTypeAheadFieldUC*. Is that intentional?

fbarthez:
No that was not intentional. AFAICT the class is not referenced, but still… Thanks, Johann!

PR #704, opened 2015-11-26
Erx framework embedding

markusstoll:
This fix automatically fixes the WebServerResources URL for embedded frameworks to be accessed via 

/WebServer/#Appname#.woa/Frameworks/...

the default all or nothing solution does not help when you only embed a part of your frameworks


fbarthez:
For me, this breaks the WebServerResources path in deployment. "MyApp.woa/Contents/Frameworks/Foo.framework" gets changed to "MyApp.woa/Frameworks/Foo.framework". Setting WOFrameworksBaseURL=/WebObjects/MyApp.woa/Contents/Frameworks has no effect.

markusstoll:
Hi, thanks for your feedback.

Do you use modified build scripts? In my build scripts (which I did not modify in this aspect) the MyApp.app bundle from the Application package contains the embedded frameworks in MyApp.woa/Contents/Frameworks, but in the WebServerResources package these are placed in MyApp.woa/Frameworks - so this behaviour is intentionally and seemed correct to me.

Could it be that you are using the MyApp-Application package for hosting the WebServerResources content? Or do we have different build scripts?

Regards, Markus

fbarthez:
Hi Markus, no modified build scripts, just standard maven. I could live with the change if at least WOFrameworksBaseURL were honored, but the way it is I have to take it all out to get a working deployment.

markusstoll:
Hi, giving WOFrameworksBaseURL higher prio might be a solution to this problem, I am just still trying to understand whats happening

fbarthez:
AFAICT on line 86 of ERXDeployedBundle, you set the base URL when the framework is embedded. Only if the framework is not embedded do you fall back to the WOFrameworksBaseURL.

markusstoll:
yes, because that was the point of this change, let me explain

with embedding only some of your frameworks, you have a problem, as WOFrameworksBaseURL can only override the location for  WebServerResources for ALL Frameworks. So my intention was to automatically take care of the path for embedded frameworks and treat installed frameworks the old way (with the option of changing the path for the latter using WOFrameworksBaseURL).

So as I see it, there are 2 options to approach this problem:
- get into details, why the WebServerResources package is different with your built
- add 2nd property (for example WOEmbeddedFrameworksBaseURL) that is able to override the path for embedded frameworks.

I would say, lets first find the root of the problem…

Markus

fbarthez:
Well, I've just created a new maven app from scratch and added some framework dependencies to it. It generates a webserver resources archive with a structure of "bar.woa/Contents/Frameworks/Ajax.framework". So this is the standard setup for everybody using maven. With maven, all frameworks are embedded by default.
I'd suggest to add a property by which to enable the special handling of embedded frameworks.

markusstoll:
Yes, if it is that way, we need a switch to toggle between these two build styles….. 
I rather would prefer to play around with this for myself.
can you supply such a mini test app with a maven build script (i have no maven experience)???

hprange:
@fbarthez did you make it work for you? I'm having the same issue after upgrading to Wonder 6.1.5.

@markusstoll looks like Wonder 6.1.5 is broken for Maven users. I see that you started a separate pull request [#727](https://github.com/wocommunity/wonder/pull/727) to handle that problem. However, it ended up closed. Am I missing something or is this still an open issue for Maven users? Is there any known workaround?

fbarthez:
I don't really remember any more… I think I just reverted the change to ERXDeployedBundle in my fork.

markusstoll:
Hi @hprange,

I recall writing to you at the time being about this issue because I was alien to maven deployment and a little clueless. However, in the mean time I learnt, that this issue is not only a problem with maven deployment. 

see this commit:
https://github.com/markusstoll/wonder/commit/2c03459b65fa66131914a615479861be3eb94ac4

which should explain the problem, unfortunately there is no automatic detecting of the proper deployment of web server resources. So my new approach would be to activate this using a property, with the property unset, the whole stuff behaves as did ever before.

If you agree, I will create a pull request from this

hprange:
Hi @markusstoll,

Sorry. I didn't see the e-mail you sent back in December. My bad.

Anyway, I tried your fix with a real application in our staging environment and I found some issues.

1) When leaving the `WOEmbeddedFrameworksPath` property unset, the base URL for embedded frameworks is `/WebObjects/my-app.woa/null`, which is wrong.

2) Frameworks aren't embedded as the result of Maven builds. I guess the same may happen for Ant builds. Why requiring users to set a property in that sense? Looks like a hack.

IMHO, the `WOEmbeddedFrameworksPath` property doesn't solve the real problem. The `ERXDeployedBundle` class determines if frameworks are embedded or not. In my scenario, the `isEmbeddedFramework` property is still set to `true` even though frameworks are not embedded. That is the real problem. And fixing it will make the solution work for you and for existing Maven users. We can fix the problem by just fixing the condition that determines if frameworks are embedded or not. I guess we have two options:

1) Fix that condition: `if(bundlePath().startsWith(NSBundle.mainBundle().bundlePath()))`.

Below are the values of bundle path in my environment.

**BundlePath**: /opt/Local/Library/WebObjects/Applications/my-app.woa/Contents/Resources/Java/my-package/my-app/my-app-core/1.0/my-app-core-1.0.jar
**MainBundlePath**: /opt/Local/Library/WebObjects/Applications/my-app.woa
**Is Embedded Framework**: true

What values do you have for bundle path in your environment? Can we use something else to determine if frameworks are embedded or not?

2) If we cannot figure out a better way to determine when frameworks are embedded or not, we can create a property to enable/disable  URL resolution for embedded frameworks.

In any case, if you want to keep the `WOEmbeddedFrameworksPath` property (which may be useful), I suggest you define a default value for this property (`Frameworks` or `Contents/Frameworks`).

markusstoll:
Hi @hprange,

thanks for your input. Before finding a fix, I need to understand the problems with maven builds, which I obviously still don't understand.

your example 

BundlePath: /opt/Local/Library/WebObjects/Applications/my-app.woa/Contents/Resources/Java/my-package/my-app/my-app-core/1.0/my-app-core-1.0.jar
MainBundlePath: /opt/Local/Library/WebObjects/Applications/my-app.woa
Is Embedded Framework: true


looks to me rather like the main bundle than a framework? With classical dev setup, the main bundle is not loaded as ERXDeployedBundle. So first I would like to understand what is happening here.

Markus

hprange:
@markusstoll `my-app-core` is a framework. Anyway, I should have selected a better example. Below is the same information for a resource inside ERExtensions.

**BundlePath**: /opt/Local/Library/WebObjects/Applications/my-app.woa/Contents/Resources/Java/wonder/core/ERExtensions/6.1.5/ERExtensions-6.1.5.jar
**MainBundlePath**: /opt/Local/Library/WebObjects/Applications/my-app.woa
**Is Embedded Framework**: true

BTW, BundlePath is the result of calling the `bundlePath()` method, and MainBundlePath is the result of calling the `NSBundle.mainBundle().bundlePath()` method on `ERXDeployedBundle` constructor.

So, `ERExtensions-6.1.5.jar` is embedded into `my-app.woa`. The web server resources included into ERExtensions are deployed in separate. They are not embedded into the main app bundle.

Maven builds produce two artifacts: the app bundle and web server resources bundle. Web server resources are deployed into Apache's document root folder. Considering that, looks like comparing the prefix of the bundle path with the main bundle path isn't enough to decide whether the resources are embedded or not.

Does it make sense?

markusstoll:
I am beginning to understand: you want to say to me: there is no such thing as fully embedded frameworks with maven builds? This does not yet make sense to me, having dependency management that takes care of embedding the right jars but not of the corresponding webserver resources…?

For my apps, I make heavy use of embedding frameworks and am really glad that they include webserver resources, so I can deploy apps that use different versions of the same framework. How ist this handled with maven builds?

fbarthez:
The web server resources are embedded into the .woa as well. But you wouldn't want to make that accessible on a web server. So you use the second build product, which is basically just a convenience.

hprange:
That reminds me of something @fbarthez mentioned earlier in that pull request. If the `WOFrameworksBaseURL` property is set, shouldn't it be honored no matter what?

markusstoll:
thanks for your input, I think I understood whats happening. I am currently testing a change with my classical deployment. After that, I would ask you to test this with maven builds, too.

markusstoll:
Hi @fbarthez and @hprange,

I agree, if you supply `WOFrameworksBaseURL`, this should be honored. However, `WOFrameworksBaseURL` is set automatically to a default value, so I have to compare against the default value.

This branch should take care of the issues properly:
[https://github.com/markusstoll/wonder/commits/ERXDeployedBundle-Fix]

I tested on a deployment with all variants of supplied Properties and it behaved correctly. Could you test this with a maven build?

Remark: In your example the framework was embedded, so flag `isEmbeddedFramework` was set correctly to `true`, but the path was not yet determined correctly due to a typo.

hprange:
@markusstoll I'll give it a try. Thanks!

hprange:
@markusstoll that solution works with a small change. https://github.com/hprange/wonder/commit/8067c4e522f0dca578389438f4980ede6355ade3 contains the fix and explains why it's required. I've tested this patch with a real application and it's working just fine.

@fbarthez would you mind to try the ERXDeployedBundle-Fix + my commit? Does it work for your projects?

Thank you guys for all your effort!

markusstoll:
I agree with your fix, thanks

hprange:
@markusstoll do you want me to create the pull request? Or would you like to have the pleasure? :)

fbarthez:
I won't be able to test this in the next days, way too much other stuff – sorry! If it works for you Henrique, it'll most likely work for me…

markusstoll:
@hprange 
Sorry, I though it was already taken care of. 
Done: [https://github.com/wocommunity/wonder/pull/781]

PR #709, opened 2015-12-02
standardize eclipse settings

darkv:
This pull request sets standardized settings for all Wonder Eclipse projects:

• uses the standard formatter 'Eclipse'
• use Unix line endings
• use UTF-8 for file encodings
• enable Java editor save actions

This is a preparation to set a standardized code formatting for Wonder which currently is quite chaotic. Before merging this pull request most other pull requests should be merged first. After that a complete reformat of the Wonder code base can be done in another pull request.


paulhoadley:
Hey @darkv! Have you got any interest in re-doing this? I like the sentiment, but now that it has merge conflicts it will need someone to adopt it.

darkv:
Hi Paul,
back then I felt having a common code formatting would greatly improve the project. Alas, nowadays we know that there won‘t be that big development effort coming to Wonder and thus I am not sure if it make sense anymore. Reformatting all code would make it more cumbersome for all that people still maintaining their own fork as they would have to reformat their code too to match and also existing PRs would need some update. That could bring that small number of PRs still tickling in to a stop.
So I would tend to close this story and move on.

paulhoadley:
I agree Johann. It was a good idea at the time.

PR #710, opened 2015-12-06
Update flying saucer to 9.0.7 to support .pdf files with vector graphics

rparada:
I am not sure why the pull request shows 5 commits. Really, the only ones that are needed are:

35631cf - Updated flying saucer libraries to 9.0.7 to support .pdf files with vector graphics
e9b21d4 - Removed jar file replaced by 3 updated jar files: flying-saucer-*.jar

The other two commits may have to do with the fact that my fork of wonder was very old and I tried to get it in sync with the current version.

The flying saucer library was updated to version 9.0.7.  It comes in three files:
flying-saucer-core-9.0.7.jar
flying-saucer-pdf-9.0.7.jar
flying-saucer-swt-9.0.7.jar

These three jars replace the core-renderer-20101006.jar file.  The updated jar files will allow the use of the img html tag to reference .pdf files that contain vector graphics.  

I checked the checked the classes in the flying saucer jar files and compared to the core-renderer-20101006.jar that they replace.  I found out that the new jars are practically a super set with only a few classes missing which may have to do with simple organization of the code.

I also checked the new flying saucer jar files with the script checkJavaVersion.sh (obtained from https://github.com/beardedN5rd/java-helper/blob/master/scripts/checkJavaVersion.sh) and I found out that they require java 1.4 which means it should be safe to merge this into the wonder6 branch as well.


rparada:
I am not sure why the pull request shows 5 commits.  Really, the only ones that are needed are:

35631cf  Updated flying saucer libraries to 9.0.7 to support .pdf files with vector graphics
e9b21d4  Removed jara file replaced by 3 updated jar files: flying-saucer-*.jar

The other two commits may have to do with the fact that my fork of wonder was very old and I tried to get it in sync with the current version before adding the 9.0.7 flying saucer jar files, updating the .classpath and removing the jar file that they replaced.

darkv:
Cherry picked the two commits into Wonder 6 and 7.

PR #713, opened 2015-12-12
When ERXWOComponentContent has a single children WODynamicElement, th…

spelletier:
Based on wonder_6 branch.

This fix situation like this:
<wo:PageContainer><wo:AjaxUpdateContainer id="xxx">
  ....content ....
</wo:AjaxUpdateContainer></wo:PageContainer>

Where the actual code skip the AjaxUpdateContainer completely.

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/components/ERXWOComponentContent.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/components/ERXWOComponentContent.java
@@ -154,7 +154,7 @@ public class ERXWOComponentContent extends WODynamicElement {
     	WOElement content =  component._childTemplate();
     	WOElement result = null;
     	String templateName = (_templateName == null) ? null : (String) _templateName.valueInComponent(component);
-    	if (content instanceof WODynamicGroup) {
+    	if (content.getClass() == WODynamicGroup.class) {
 			WODynamicGroup group = (WODynamicGroup) content;
 			if (templateName == null) {
 				// MS: If you don't set a template name, then let's construct all the children of 
@@ -190,7 +190,7 @@ public class ERXWOComponentContent extends WODynamicElement {
     		if(name.equals(templateName)) {
     			result = template;
     		}
-		} else if (content instanceof WOHTMLBareString && templateName == null) {
+		} else if (templateName == null) {
 			result=content;
 		}
     	return result;


darkv:
Hi Samuel,

I cannot reproduce the problem you want to fix with this pull request neither on wonder_6 nor master branch. Can you elaborate more on that topic?

spelletier:
You need to use <wo:ERXWOComponentContent /> in the PageContainer as the content placeholder to experience the problem. If there is a space, a return or anything else before the opening or after the closing tag of the content the problem does not occur.

For example, if your component is like this:
<wo:PageContainer>
   <wo:AjaxUpdateContainer id="xxx">
      ....content ....
   </wo:AjaxUpdateContainer>
</wo:PageContainer>

or this:
<wo:PageContainer><wo:AjaxUpdateContainer id="xxx">
....content ....
</wo:AjaxUpdateContainer>
</wo:PageContainer>

or this:
<wo:PageContainer> <wo:AjaxUpdateContainer id="xxx">
....content ....
</wo:AjaxUpdateContainer></wo:PageContainer>

You won't experience the problem.

In this example, a div with the specified id is supposed to be present to create the updatable zone but it is not present without my fix.

I think there is something similar in the AjaxModalDialog, I saw a warning about adding a space before or after a form when it is the first element inside the dialog last week.

chillatgvc:
Yes I remember hitting this, from AMD:
// WOForm expects that it is inside a WODynamicGroup and relies on WODynamicGroup having setup the WOContext to correctly 
// generate the URL.  It should not (IMO), but it does.  If you have anything before the WebObject tag for the form 
// (a space, a carriage return, text, HTML tags, anything at all), then the WO parser creates a WODynamicGroup to hold that.  
// If  the WebObject tag for the form is the very first thing in the template, then the WODynamicGroup is not created and 
// invalid URLs are generated rendering the submit controls non-functional.  Throw an exception so the developer knows what is 
// wrong and can correct it.
if (_actionResults != null && (_actionResults.template() instanceof WOForm ||
  		                       _actionResults.template() instanceof ERXWOForm)) {
	throw new RuntimeException(_actionResults.name() + " is used as contents of AjaxModalDialog, but starts with WOForm tag.  " +
			"Action elements inside the dialog will not function.  Add a space at the start or end of " + _actionResults.name() + ".html");
}

spelletier:
This is a totally different problem.

darkv:
Suggested fix worked on a test project. Thanks!

PR #719, opened 2015-12-28
Prevent NPE when first array is null

fbarthez:
For some reason I don't understand, the new implementation introduced in #697 throws an NPE, while the previous one did not.


darkv:
The change seems due to passing a Collection instead of an NSArray to the constructor of NSMutableArray. With an NSArray param that class handles a null value gracefully whereas in the case of a Collection, List or Vector a NPE is thrown. That seems not very logical and should be changed to support null values too. I will create a patch for that.

darkv:
Commit 44d56b1 should be fixing your problem, can you give it a try?

PR #721, opened 2016-01-06
Fix handling of owned relationships in ERModern

fbarthez:
This fixes an issue with deletion of an owned relationships, which would cause an NPE in EOEditingContext.
Also, the "query related" and "remove related" buttons will never be shown for owned relationships.


Frameworks/D2W/ERModernDirectToWeb/Sources/er/modern/directtoweb/components/buttons/ERMDRemoveRelatedButton.java
```diff
@@ -191,12 +194,14 @@ public Boolean showRemoveButton() {
...
     			EORelationship relationship = masterEntity.relationshipNamed(dds.detailKey());
     			EORelationship reverseRelationship = relationship.inverseRelationship();
-    			if(isRemoveable) {
-    				if(reverseRelationship == null) {
-    					_showRemoveButton = Boolean.TRUE;
-    				} else {
-    					_showRemoveButton = !reverseRelationship.isMandatory();
-    				}
+    			if(isRemoveable && !relationship.ownsDestination()) {
+                    if (!relationship.ownsDestination()) {
```
darkv: What purpose has this check? This will always evaluate to true as in the line before it is already part of the if-condition.

fbarthez: None at all. Thanks!

PR #727, opened 2016-02-03
Fix framework embedding for Maven builds

markusstoll:
This is a proposal to solve this problem:

I was unaware that using Maven for building your application, the WebServerResources bundle has a different layout (embedded framework stuff is in Contents/Frameworks rather than in Frameworks folder).

So I added an automatic detection for the build style and switches to override.

If there is a better way to solve this, let me know, I would gladly rework this


markusstoll:
There was a misunderstanding, the deployment schemes did not differ due to maven / ant but rather due to different ant scripts, obviously there are 2 different templates in the wild

In https://wiki.wocommunity.org/display/WOL/Alternative+Ant+Build+Script+for+Fully+Embedded+and+Split+Install+Bundles?src=search

frameworksBaseURL=/WebObjects/${build.app.name}.woa/Frameworks

is referred, however some people seem to use an ant script that would require 

frameworksBaseURL=/WebObjects/${build.app.name}.woa/Contents/Frameworks

have to find another solution

PR #731, opened 2016-04-14
AjaxProxy Enhancement

lbane:
There seems to be a confusion between the binding names "AjaxBridge" and "JSONRPCBridge" in AjaxProxy. Both binding names are used, but I believe they should mean the same object. The patch changes AjaxProxy to accept both names.

The patch also provides some enhancements to the documentation, i.e. a warning to never omit the proxy object.

### DIFF
--- a/Frameworks/Ajax/Ajax/Sources/er/ajax/AjaxProxy.java
+++ b/Frameworks/Ajax/Ajax/Sources/er/ajax/AjaxProxy.java
@@ -20,16 +20,18 @@ import er.extensions.appserver.ERXResponseRewriter;
 import er.extensions.appserver.ERXWOContext;
 
 /**
- * Handles javascript-java communication (client-server) between the javascript world running in a web browser and the
+ * A RPC mechanism to call methods on the Java server side from a JavaScript client.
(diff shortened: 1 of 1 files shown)


paulhoadley:
Hi Ralf,

I've just started seeing some output from your log warning:

    log.warn("No proxy binding given, so using parent component. This is probably a very bad idea.");

I tracked it down to `AjaxFlexibleFileUpload`, which I'm using from `ERAttachmentFlexibleUpload`. I notice that `AjaxFlexibleFileUpload` certainly doesn't set the `proxy` binding:

    AjaxProxy : AjaxProxy {
      name = ajaxProxyName;
      proxyName = "wopage";
    }

How dangerous is this? The component certainly still _works_, but is this exposing a significant security risk, and if so can we fix it in `AjaxFlexibleFileUpload`?

lbane:
AjaxProxy currently publishes **every** public method of the object on the client side. For components this means for example `WOComponent.valueForKeyPath` and `WOComponent.takeValueForKeyPath` are directly callable by the client. A malicious user may call `valueForKeyPath("application.terminate")` or every other path reachable by session or application. I call this a serious issue.

PR #734, opened 2016-04-16
fixes arrayMinusObject if object is contained more than once

darkv:
The method List.remove(Object) which replaced the previous NSMutableArray.removeObject(Object) only removes the first match from the object in contrast to *all* occurrences as removeObject does. This fix keeps the List method signature but repeats the remove until no removal happens to match the behavior of the NS-class method.


paulhoadley:
Works for me: this restores the behaviour of the previous implementation with respect to duplicate objects.

PR #737, opened 2016-04-19
Fixes YUI.framework for Wonder 7.

paulhoadley:
Although it's gathering dust in the `Archive` directory, we still have an ancient app dependent on `YUI.framework`. (It seemed easier to fix the framework than extricate it from the app.) Several classes were calling the now non-existent `AjaxUtils.stringValueForBinding()`, so here we just substitute the alternative `ERXComponentUtilities.stringValueForBinding()`.


paulhoadley:
Ah crap, this includes a bunch of extraneous merge commits. Will fix.

PR #740, opened 2016-04-21
Adds log warning to parseTemplatedStringWithObject()

paulhoadley:
Adds a simple log warning to the case where this method would otherwise fail silently.

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXSimpleTemplateParser.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXSimpleTemplateParser.java
@@ -35,7 +35,7 @@ public class ERXSimpleTemplateParser {
     public static final String DEFAULT_DELIMITER = "@@";
 
     /** logging support */
-    private static final Logger log = LoggerFactory.getLogger(ERXSimpleTemplateParser.class);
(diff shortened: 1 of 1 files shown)


Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXSimpleTemplateParser.java
```diff
@@ -35,7 +35,7 @@
     public static final String DEFAULT_DELIMITER = "@@";
 
     /** logging support */
-    private static final Logger log = LoggerFactory.getLogger(ERXSimpleTemplateParser.class);
+    private static final Logger LOG = LoggerFactory.getLogger(ERXSimpleTemplateParser.class);
```
darkv: I think we should keep the character case of that variable. Surely it is quite common to make static variables uppercase but in the case of the _log_ variable we have it lowercase in all Wonder source.

paulhoadley:
Fine by me. I think we should uppercase it everywhere we see it, but it's not something I feel strongly enough about to defend at any length. I'll adjust the pull request.

spelletier:
All caps are supposed to be constants. A logger is not a constant, it is a reference to shared active object. I've done some google and it seems to be the preferred way to reserve all caps to constant only.

Personally, I dislike the underscore and all caps so I use camel Caps for my constant names like Switft suggest.

PR #745, opened 2016-05-18
Add contains any of these objects in ERXKey

spelletier:


paulhoadley:
Hi Samuel,

Just having a look at this. Is "May 56" a typo?

`* @since May 56, 2016`

paulhoadley:
Hi Samuel,

I'm having a look at `containsAnyOfTheseObjects()`. Recall from my example on the mailing list that I have two entities, `User` and `Notice`, each of which has a to-many relationship to `Role` called `roles`, and `Role` has _neither_ of the back-relationships to those entities. The business rule I want to model is finding `Notice` objects where `Notice.roles` contains _any_ of the objects in some `User.roles`. While this works:

    NSMutableArray<EOQualifier> roleQuals = new NSMutableArray<EOQualifier>();
    for (Role r : user.roles()) {
        roleQuals.add(Notice.ROLES.containsObject(r));
    }
    EOQualifier roleQual = ERXQ.or(roleQuals);

I had hoped that this would be a more elegant shortcut:

    EOQualifier roleQual = Notice.ROLES.containsAnyOfTheseObjects(user.roles());

Are there any other constraints on `containsAnyOfTheseObjects()` that are not mentioned in your Javadocs? Because while this works in-memory (specifically, in WOUnit testing), as soon as I head to the database (PostgreSQL), I get an NPE from `ERXExistsQualifier$ExistsQualifierSQLGenerationSupport.sqlStringForSQLExpression()`. Presumably this is because when that method is called, its `expression` argument is `<com.webobjects.jdbcadaptor.PostgresqlExpression: "null" withBindings: >`.

    NullPointerException
      at er.extensions.eof.qualifiers.ERXExistsQualifier$ExistsQualifierSQLGenerationSupport.sqlStringForSQLExpression(ERXExistsQualifier.java:270)
      at com.webobjects.eoaccess.EOQualifierSQLGeneration$Support._sqlStringForSQLExpression(EOQualifierSQLGeneration.java:165)
      at com.webobjects.eoaccess.EOSQLExpression.prepareSelectExpressionWithAttributes(EOSQLExpression.java:997)
      at com.webobjects.jdbcadaptor.JDBCExpression.prepareSelectExpressionWithAttributes(JDBCExpression.java:146)
      at com.webobjects.jdbcadaptor.PostgresqlExpression.prepareSelectExpressionWithAttributes(PostgresqlExpression.java:770)
      at com.webobjects.eoaccess.EOSQLExpressionFactory.selectStatementForAttributes(EOSQLExpressionFactory.java:225)
         ... skipped 6 stack elements
      at com.webobjects.eocontrol.EOObjectStoreCoordinator.objectsWithFetchSpecification(EOObjectStoreCoordinator.java:488)
      at com.webobjects.eocontrol.EOEditingContext.objectsWithFetchSpecification(EOEditingContext.java:4069)
      at er.extensions.eof.ERXEC.objectsWithFetchSpecification(ERXEC.java:1301)
      at com.webobjects.eocontrol.EOEditingContext.objectsWithFetchSpecification(EOEditingContext.java:4444)
      at er.extensions.eof.ERXFetchSpecification.fetchObjects(ERXFetchSpecification.java:155)
      at net.logicsquad.notice.model._Notice.fetchNotices(_Notice.java:426)

That is, I can see that `containsAnyOfTheseObjects()` works in memory, but not in SQL. Any thoughts?

spelletier:
Hi Paul,

This remind me something in the Aaron presentation about the ERXExistsQualifier not supporting flattened many to many relationships SQL generation. I suppose this is how your  `User` -> `Roles` is implemented...

I will try to understand better what is going out, I think this is a real life situation without a good solution right now.

BTW, I fixed the dates.

spelletier:
I added the SQL generation support for Many to Many. Aaron will be happy too I think.

paulhoadley:
> This remind me something in the Aaron presentation about the ERXExistsQualifier not supporting flattened many to many relationships SQL generation.

Ah right. Yes, `User.roles` and `Notice.roles` are both flattened. I had forgotten about that limitation.

> I added the SQL generation support for Many to Many. Aaron will be happy too I think.

Nice work! This certainly seems to work for me. Should we get some more eyes on it? @recurve would you mind taking a look at this?

recurve:
That’s great Paul that flattened relationship support was added! :-)

The best way to check would be run it and see under different scenarios.

Unit tests would be great but this requires setting up and connecting to a DB and generating SQL, many pieces in the air.

I’m sorry but I can’t look at the improved code right now but would be fun to do in Montréal… but from experience, you need physically test the SQL generation and flex it to really know. It’s generally hard to eyeball.
AARON ROSENZWEIG / Chat 'n Bike
e:  aaron@chatnbike.com  t:  (301) 956-2319

spelletier:
I had to look at the SQL to understand with was going on and made the thing work. This is how I found a problem with the attributed used for the destination unity in the original code that assumed the relationship pointed to the primary key instead of using the destination attributes of the relationship.

I tested with an app using a N-N by adding a fetch similar to the one Paul need using the EXISTS and IN version of the qualifier. I also tested an existing fetch in another app.

The risk is minimal, it is a feature addition and I am very confident id this code.

recurve:
That sounds great Samuel.

I’m not one of the people that can incorporate a pull request - hopefully someone will do that soon :-)

Thank you for doing that Samuel. It’s very helpful. 
AARON ROSENZWEIG / Chat 'n Bike
e:  aaron@chatnbike.com  t:  (301) 956-2319

paulhoadley:
> The risk is minimal, it is a feature addition and I am very confident id this code.

Sounds good Samuel. I just want to get some more eyes on it before I hit merge. Johann will take a look in the next few days. As hard as it is with the developer pool shrinking, I think we should aim to get changes reviewed by multiple people when we can.

darkv:
Looks good for me. Perhaps you could add a varargs variant of containsAnyOfTheseObjects(NSArray)?

spelletier:
I decided to not add it because I do not think it would be often used, we rarely have EO in contants in our code. If the list EOs is built from UI or saved in an EO, it will already be inside an NSArray.

PR #746, opened 2016-05-24
Fix the generic types in operators that works on array.

spelletier:
The type of the new key remains the same, it is known. This bothered me for a while but I use many of these in my current projet so I decided to fix the class.

Personally, I would remove all the versions that have the "next key" as param, I think they pollute the class. This is especially true for the atXXX(ERXKey<U> key). Is there anybody using these ?

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/eof/ERXKey.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/eof/ERXKey.java
@@ -338,17 +338,14 @@ public class ERXKey<T> {
 	 * Creates a new ERXKey that appends ERXArrayUtilities'
 	 * {@code @flatten} operator to this key.
 	 * 
-	 * @param <U>
-	 *            the type of the next key
-	 * 
 	 * @return an {@code ERXKey<U>} wrapping the {@code thisKey.@flatten}
 	 *         keypath
 	 * 
 	 * @see er.extensions.foundation.ERXArrayUtilities.FlattenOperator
 	 *      FlattenOperator
 	 */
-	public <U> ERXKey<U> atFlatten() {
-		return (ERXKey<U>) append(ERXKey.flatten());
+	public ERXKey<T> atFlatten() {
+		return append(ERXKey.flatten());
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


fbarthez:
This broke the build, on the [community jenkins](https://jenkins.wocommunity.org/job/WonderMvnBuild/66/wonder.tests$ERXTest/console), on travis and on my machine. I haven't looked into it, so I don't know whether this is a problem with ERXTest or whether something is actually amiss…

spelletier:
ant tests.all fail on my machine with something else so I never see that problem.
failure: test_parse_guess_r(er.chronic.ParserTest$ParserInnerTest): expected:<Thu Aug 17 13:00:00 PDT 2006> but was:<Wed Aug 16 13:00:00 PDT 2006>
...

I fixed the ERXKey tests, they compile now they should pass now. I created a new pull request.

Maybe someone can help me get the tests working on my dev machine during WOWODC.

fbarthez:
You're not alone: http://lists.apple.com/archives/webobjects-dev/2016/Jan/msg00009.html
And: http://wonder-disc.narkive.com/Lw59TD1G/really-strange-ec-test-failure

spelletier:
I found a way to fix my error...It is in ParserTest. 

The test try to run in 2 time zone, the current system time zone and America/Los_Angeles. When the tests try to run in another timezone than the system one, they fail, probably there is something that does not uses the default time zone or cache it. The offset between the expected time and computed are always 3h and this correspond to the offset between America/Los_Angeles and my current time zone.

If I remove the run with the fixed America/Los_Angeles time zone, no more errors.

fbarthez:
Yep, that's why I [added timezone arguments to the VM launch](https://github.com/fbarthez/wonder/commit/78851ba01a3f282e17537aa3ef11c312994a5c35#diff-77a97b88c87ff70bed44c15f528b386dR1257) when I looked at it.

How about the testPlainOneThreadAutoLockingCoalesceInRequest one, does that pass for you?

spelletier:
I fixed the ERChronic tests, the time reference was a constant but needed to be adjusted for each time zone. I will create a pill request with the fixe.

Now, all tests pass on my machine.
ant tests
ant tests.run
...
     [java] tests run: 840
     [java] tests FAILED: 0
     [java] time elapsed: 20 sec
 
Running on a Mac with OS X 10.11.5, jdk 8u92 and Eclipse 4.4.2

PR #749, opened 2016-05-26
Add uuid support

spelletier:
Add support for auto generated UUID type 4 (random) as primary key.
Add a uuid prototype for FrontBase, MySQL and PostgreSQL. I do not have instance of other type of database to test them.
Add support for UUID primary key in ERRest.

### DIFF
--- a/Frameworks/Core/ERPrototypes/Resources/erprototypes.eomodeld/EOJDBCPostgresqlPrototypes.plist
+++ b/Frameworks/Core/ERPrototypes/Resources/erprototypes.eomodeld/EOJDBCPostgresqlPrototypes.plist
@@ -359,6 +359,13 @@
             valueClassName = NSString; 
             valueType = S; 
             width = 10000000; 
+        }, 
+        {
+            columnName = uuid; 
+            externalType = bytea; 
+            name = uuid; 
+            valueClassName = NSData; 
+            width = 16; 
         }
     ); 
     attributesUsedForLocking = (boolean, id, jodaDateTime); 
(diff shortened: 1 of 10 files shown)


Frameworks/Core/ERPrototypes/Resources/erprototypes.eomodeld/EOJDBCPostgresqlPrototypes.plist
```diff
@@ -359,6 +359,13 @@
             valueClassName = NSString; 
             valueType = S; 
             width = 10000000; 
+        }, 
+        {
+            columnName = uuid; 
+            externalType = bytea;
```
paulhoadley: Hi Samuel,

Would it be possible to use [PostgreSQL's `uuid`](https://www.postgresql.org/docs/current/static/datatype-uuid.html) as the external type here? Or does it create problems elsewhere?

spelletier: Paul,

PG UUID seems to map to Java UUID and is a very PG specific. I do not remember having tried to map a UUID column to an NSData with EOF.

Keep in mind that EOF only supports Integer, Long, String, Date and NSData as primary key type.

Samuel

paulhoadley: Ah, thanks Samuel—it was lost on me that your overall aim was primary key support.

PR #754, opened 2016-06-01
New ERAjaxTagField tag editing component

fbarthez:
New ERAjaxTagField component that provides an alternative GUI to the existing ERTagField. It's only a wrapper around CCTagEditor, which in turn is a WO-wrapper around Nicolás Bevacqua's Insignia JS tag editing module and his Horsey auto-completion module.
Also adds another thin wrapper, ERD2WEditTags, for use with D2W.

This adds a dependency on ERCoolComponents to ERTaggable. Alternatively, CCTagEditor could be moved into ERTaggable, which would get rid of the dependency on ERCoolComponents but add another dependency on the Ajax framework.

erd2wedittags


fbarthez:
In it's current state, this causes the maven build to run forever. I've found the problem and have a fix that I'll push later today.

fbarthez:
Rebasing this is too much of a hassle. I'm closing it and creating a new PR.

PR #766, opened 2016-06-13
update some libs in ERJars

darkv:
This updates httpcore and ognl libraries to their current versions. Those are maintenance release and are fully backward / binary compatible.


maiksd:
I get

> BUILD FAILED
> /Users/maik/src/git/wonder/build.xml:12: The following error occurred while executing this line:
> /Users/maik/src/git/wonder/Build/build/build.xml:1153: The following error occurred while executing this line:
> /Users/maik/src/git/wonder/Build/build/build.xml:40: Please check versions of jar files included in ERJars project. See properties in Build/build//build.xml.

after pulling this, but can't spot the problem really...

darkv:
Should be fixed now. Forgot that those libs are referenced in the build.xml and pom.xml files…

PR #767, opened 2016-06-16
deprecate ERXFile

darkv:
As noted in the JavaDocs this class seems to be obsolete as of

> WO 5.4.3, Mac OS X 10.5.8, Java 1.5.0_22

Retested with WO 5.4.3, Mac OS X 10.11.5, Java 1.8.0_92 and still described problem that ERXFile tries to solve does not appear anymore.

This deprecates the class and replaces single usage in ERXFileUtilities by plain java.io.File. As this change is for Wonder 7 only we could altogether remove that class. Any thoughts?


paulhoadley:
> Any thoughts?

Fine by me, don't think I've ever used it.

darkv:
So let's ditch that class.

PR #771, opened 2016-06-20
Adds mod_proxy support

pascalrobert:
Last year, a page in JavaMonitor was added to help migrate to mod-proxy. That page have code to better support the load balancing of mod-proxy, so this is the code added to ERXApplication so that everyone can benefit of the mod-proxy support.


darkv:
Hi Pascal, you want this to be merged into Wonder 6 too?

pascalrobert:
Why not, so yes.

Issue #774, opened 2016-06-27
Occasional NPEs from ERXTimeZoneDetector


paulhoadley:
I started a [mailing list thread](http://prod.lists.apple.com/archives/webobjects-dev/2016/Apr/msg00081.html) on this issue back in April. I’m seeing occasional NPEs from `ERXTimeZoneDetector` like this:

    java.lang.NullPointerException: timeZone must not be set to null
        at er.extensions.appserver.ERXSession.setTimeZone(ERXSession.java:810)
        at er.extensions.components.ERXTimeZoneDetector.takeValuesFromRequest(ERXTimeZoneDetector.java:148)
        at com.webobjects.appserver._private.WOComponentReference.takeValuesFromRequest(WOComponentReference.java:119)

`ERXSession.setTimeZone()` doesn’t like nulls, and that’s fair enough. It’s being called from here:

    public void takeValuesFromRequest(WORequest request, WOContext context) {
        super.takeValuesFromRequest(request, context);
        if (shouldPostData() && request.formValueForKey(TIMEZONE_DATA_KEY) != null) {
            ...
            TimeZone tz = zoneWithRawOffset(rawOffset, dst, southern);
            session.setTimeZone(tz);
        }
    }

That is, when `ERXTimeZoneDetector` thinks it’s found the `TimeZone`, it calls `setTimeZone()`, except sometimes `tz` is `null`. In the resulting log output, `ERXBasicBrowser` is suggesting the culprit is usually IE 8, sometimes IE 9 and IE 7. I don’t know whether this is a legitimate app user, or someone browsing the login page, or a bot.

It turns out that `detectedTimeZone` is consistently “180,1,0”. It’s the DST parameter that’s causing the problem. `zonesWithRawOffset(180, true, false)` returns `null`, but `zonesWithRawOffset(180, false, false)` returns an array of 27 zones. This is using a recent Java 8 release, so presumably the timezones should be quite up to date.

Again, I don't know what the root cause is here (real user with misconfigured user agent, misconfigured bot, real timezone that Java doesn't know about, bug in the detection Javascript...), but in any case I would argue that it's a bug for `ERXTimeZoneDetector` to call `setTimeZone(tz)` when it knows `tz` is `null`. I propose that `ERXTimeZoneDetector` guards its call to `setTimeZone()`with a check for `null`. and logs something informational if it's unable to detect the timezone. I'll submit a pull request soon.

paulhoadley:
Fixed by #775.

PR #775, opened 2016-06-27
Prevents ERXTimeZoneDetector calling setTimeZone() with null argument

paulhoadley:
As discussed in #774, `ERXTimeZoneDetector` will call `ERXSession.setTimeZone()` with a `null` argument if it can't find a timezone, causing an NPE. Here we just add a check for `null`, and log a warning (and do nothing else) if found.


darkv:
Thanks :)

PR #776, opened 2016-06-27
Small fixes

darkv:
This patch contains some smaller fixes:

- passing _null_ to constructors `NSSet(NSArray)` and `NSSet(NSSet)` should throw an `IllegalArgumentException` instead of running into an NPE
- remove some association vars that are already defined in parent class
- do not throw away coerced value from `_validateValueForKey` (used by partials)
- `isInet4IPAddressWithinRange` did not check correct param


darkv:
Glad that everything works as expected :)

PR #785, opened 2016-07-14
ProcessChangesQueue.stop() shouldn't throw IllegalStateException

paulhoadley:
`ProcessChangesQueue.stop()` is called when `ApplicationWillTerminateNotification` is posted. If `_queueThread` is `null` or not alive, throwing an `IllegalStateException` here just prevents the application from terminating.


paulhoadley:
I'm more than happy to have this critiqued by anyone who uses `ERXObjectStoreCoordinatorSynchronizer` (or anyone at all). As noted elsewhere, Ted Archibald and I have seen this in the context of using `ERJGroupsSynchronizer.framework`, but it's not related to JGroups shutdown (which does its own thing using `ERXShutdownHook`). For what it's worth, @chillatgvc can't see a good reason to be throwing an exception here either:

> That does not seem to be a sensible reason/place to throw an exception.

PR #787, opened 2016-07-18
Add restricted choice pop-up and relationship filtering to ERMD2WListFilter

fbarthez:
Allows filtering based on the selection in a pop-up menu, e.g. for category or status filtering.
Enables use of ERMD2WListFilter in relationship edit and inspect configurations.


fbarthez:
On my local branch, I cherry-picked an additional commit from another branch, then pushed to my remote github branch. Alas, the commit doesn't show up here as I would have expected. Guess I'll just do another PR?

fbarthez:
Well, pushing to the right branch does help! This should be fine to merge now.

PR #788, opened 2016-08-02
add ERXKey to NSDictionary/NSMutableDictionary KVC

omtay:
NSDictionary valueForKey, takeValueForKey using ERXKey; NSMutableDictionary takeValueForKey using ERXKey


avendasora:
Nothing to see here. Move along. This is what happens when you start accepting how Past Dave™ did things without questioning the living crap out of it.

PR #790, opened 2016-08-02
Add support of SameSite to WOCookie

darkv:
This patch adds the SameSite support to WOCookie.

Full info on what this cookie flag does can be found at https://tools.ietf.org/html/draft-west-first-party-cookies-07 or less technical but better readable at https://www.sjoerdlangkemper.nl/2016/04/14/preventing-csrf-with-samesite-cookie-attribute/. Currently this flag is supported in Chrome 51 and Opera 39.

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/com/webobjects/appserver/WOCookie.java
+++ b/Frameworks/Core/ERExtensions/Sources/com/webobjects/appserver/WOCookie.java
@@ -25,6 +25,7 @@ public class WOCookie implements NSKeyValueCoding, com.webobjects.foundation.NSK
 	NSTimestamp _expires;
 	int _timeout;
 	boolean _isHttpOnly;
+	SameSite _sameSite;
 	@Deprecated
 	static final SimpleDateFormat TheDateFormat;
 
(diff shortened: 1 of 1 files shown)


paulhoadley:
This is very cool Johann.

Say you have an app that stores session ID in cookies:

1. Is this a viable defence against CSRF?
2. How would you get WebObjects to use `SameSite.STRICT` for the `wosid` cookie?

darkv:
Hi Paul,

1. it helps against CSRF but does of course not solve it completely
2. you could overwrite the method `_appendCookieToResponse` in your session class and modify the wosid cookie there, should be the easiest way

We could of course introduce a new property and set that value accordingly within Wonder classes automagically.

paulhoadley:
> We could of course introduce a new property and set that value accordingly within Wonder classes automagically.

I think we should definitely do this. It would seem fairly analogous with the handling of the `useSecureSessionCookies` and `useHttpOnlySessionCookies` in `ERXSession` (which I didn't even know existed until five minutes ago). We could add one more property, default to existing behaviour, and add another if-block in `_appendCookieToResponse()`.

I'll add this to my to-do list, unless you've already done it while I've been typing!

darkv:
> I'll add this to my to-do list, unless you've already done it while I've been typing!

Now that you say it, I actually did ;-) see #799

PR #791, opened 2016-08-05
Correct validation if values are coerced

darkv:
If you are coercing values in your validation methods these value changes could get lost or based on wrong values in some edge cases.
There are three main locations for key related validation:

1. EO class
2. Entity userInfo
3. Partial class

Point 3 could theoretically contain multiple partials even though this should be quite unusual. During 1. and 3. the validated value can be coerced though some of the methods did get the original value instead of the coerced one. This is inconsistent and can lead to data loss which this patch fixes.


paulhoadley:
Looks good to me Johann. I'll merge it tomorrow if there's no further input from anyone.

PR #794, opened 2016-08-15
Added multiple file selection button to ERDragAndDropUpload, with the…

frelin:
… option to show a button or use the ID from an external button. The component can also be used without content to just show a button. Included is a fix from Michael Kondratov for the completeAllFunction in dndupload.js. completeAllFunction now works as advertised.


frelin:
I have encoutered a possible incompability with IE. Let me test more thorougly before accepting the pull-request.

frelin:
All clear! The problem on the deployed site proved to be with google maps :) Feel free to accept!

PR #798, opened 2016-09-09
Add missing dependency that was introduced in beea23f

fbarthez:


pom.xml
```diff
@@ -678,6 +678,11 @@
                 <version>3.14</version>
             </dependency>
             <dependency>
+                <groupId>org.apache.poi</groupId>
+                <artifactId>poi-ooxml</artifactId>
+                <version>3.9</version>
```
darkv: Should that be 3.14? The referenced commit uses the newer version.

PR #799, opened 2016-09-09
make SameSite setting for cookies configurable

darkv:
By property *er.extensions.ERXSession.cookies.SameSite* you can define if and which SameSite setting should be used for your session and instance cookies.


paulhoadley:
I'm seeing `_setCookieSameSite()`, but I'm not seeing it called anywhere. Further, I'm not seeing `ERXProperties.enumValueForKey()`. What am I missing here?

darkv:
That happens when doing too many things in parallel. Had that method in another fork.

paulhoadley:
Looking good! I'll test it out a bit and merge tomorrow.

PR #800, opened 2016-09-12
Code cleanup

darkv:
This brings some clean up. The try-with-resource will guarantee that resources are correctly freed.

### DIFF
--- a/Frameworks/Ajax/Ajax/Sources/er/ajax/AjaxFileUploadRequestHandler.java
+++ b/Frameworks/Ajax/Ajax/Sources/er/ajax/AjaxFileUploadRequestHandler.java
@@ -121,13 +121,8 @@ public class AjaxFileUploadRequestHandler extends WORequestHandler {
 							progress.dispose();
 							throw e;
 						}
-						FileOutputStream fos = new FileOutputStream(progress.tempFile());
-						try {
+						try (FileOutputStream fos = new FileOutputStream(progress.tempFile())) {
 							progress.copyAndTrack(uploadInputStream, fos, _maxUploadSize);
-						}
-						finally {
-							fos.flush();
-							fos.close();
 						}
 						if (!progress.isCanceled() && !progress.shouldReset()) {
 							downloadFinished(progress);
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXCompressionUtilities.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXCompressionUtilities.java
@@ -33,20 +33,10 @@ public class ERXCompressionUtilities {
 	 * @return gzipped NSData
 	 */
 	public static NSData gzipInputStreamAsNSData(InputStream input, int length) {
-		try {
-			ERXRefByteArrayOutputStream bos = new ERXRefByteArrayOutputStream(length);
+		try (ERXRefByteArrayOutputStream bos = new ERXRefByteArrayOutputStream(length)) {
 			if (input != null) {
-				GZIPOutputStream out = new GZIPOutputStream(bos);
-				try {
+				try (GZIPOutputStream out = new GZIPOutputStream(bos)) {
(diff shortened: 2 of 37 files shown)


Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXFileUtilities.java
```diff
@@ -1050,77 +1024,22 @@ public static void copyFileToFile(File srcFile, File dstFile, boolean deleteOrig
...
-                			thrownException = t;
-                		}
-                    } finally {
-                    	srcChannel.close();
-                	}
-                } catch (Throwable t) {
-                	if (thrownException == null) {
-                		thrownException = t;
-                	}
+                try (FileInputStream in = new FileInputStream(srcFile);
```
paulhoadley: This is particularly nice.

paulhoadley:
Nice work Johann. I can have a scan over everything and merge tomorrow.

darkv:
That would be great! Added some more minor changes.

Frameworks/Ajax/Ajax/Sources/er/ajax/AjaxFileUploadRequestHandler.java
```diff
@@ -121,14 +121,9 @@ else if (formData.isFileUpload()) {
 							progress.dispose();
 							throw e;
 						}
-						FileOutputStream fos = new FileOutputStream(progress.tempFile());
-						try {
+						try (FileOutputStream fos = new FileOutputStream(progress.tempFile())) {
 							progress.copyAndTrack(uploadInputStream, fos, _maxUploadSize);
 						}
-						finally {
-							fos.flush();
```
paulhoadley: To be honest, the try-with-resources construct had passed me by until your commits here, so I'm no expert. It would seem, though, that it only guarantees `close()` will be called, and whether `close()` calls `flush()` is class-dependent. I'm not seeing a mention of `close()` calling `flush()` in the Javadocs for `FileOutputStream` or `OutputStream`. So, (a) do you know if it does, and (b) do you think it matters if it doesn't, or should we explicitly call `flush()` before the try-block ends?

Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXCompressionUtilities.java
```diff
@@ -33,21 +33,11 @@
...
+		try (ERXRefByteArrayOutputStream bos = new ERXRefByteArrayOutputStream(length)) {
 			if (input != null) {
-				GZIPOutputStream out = new GZIPOutputStream(bos);
-				try {
+				try (GZIPOutputStream out = new GZIPOutputStream(bos)) {
 					ERXFileUtilities.writeInputStreamToOutputStream(input, true, out, false);
 				}
-				finally {
-					try {
-						out.finish();
```
paulhoadley: As above, are we discarding the call to `finish()` here? Does it matter?

paulhoadley:
What's the benefit of `ThreadLocalRandom` over `Random`?

Frameworks/Ajax/Ajax/Sources/er/ajax/AjaxFileUploadRequestHandler.java
darkv: The flush method of FileOutputStream is empty so it does nothing. Additionally at that point the stream is not used anymore so even if there was something to flush it would have no influence.

Frameworks/Core/ERExtensions/Sources/er/extensions/foundation/ERXCompressionUtilities.java
darkv: Finish() is called from DeflaterOutputStream.close() so it is already called automatically.

darkv:
> What's the benefit of ThreadLocalRandom over Random?

The usage of `ThreadLocalRandom` over `Random` is often recommended as in a multithreaded environment it has quite better performance. It gives you an object per thread so it does not need any locking / atomic strategies which reduces the overhead.

PR #801, opened 2016-09-16
Code cleanup

darkv:
This cleans up code a lot by use of diamond operators. Some generics were added too and some private variables have been changed to be declared with interface _Map_ or _Set_ instead of specific classes.


darkv:
I don't see any problematic changes. The changes are mostly on private elements. The only public change is `ERXKeyFilter.sortOrderings()` which returns now `NSArray<EOSortOrdering>` instead of the raw type `NSArray` and `ERXKeyFilter.setSortOrderings(NSArray<EOSortOrdering>)` which previously took the raw type `NSArray` as param. Nothing which should blow up any code.

PR #802, opened 2016-09-16
New ERAjaxTagField tag editing component

fbarthez:
New ERAjaxTagField component that provides an alternative GUI to the existing ERTagField. It's only a wrapper around CCTagEditor, which in turn is a WO-wrapper around Nicolás Bevacqua's Insignia JS tag editing module and his Horsey auto-completion module.
Also adds another thin wrapper, ERD2WEditTags, for use with D2W.

This adds a dependency on ERCoolComponents to ERTaggable. Alternatively, CCTagEditor could be moved into ERTaggable, which would get rid of the dependency on ERCoolComponents but add another dependency on the Ajax framework.

edittags


paulhoadley:
Given that this just adds a component, albeit with an additional dependency for ERTaggable.framework, I suggest we merge this. Unfortunately, I cannot get ERModernMoviesDemo to run at the moment to take a look. But the code looks clean and well-documented. I vote for a merge.

PR #804, opened 2016-09-16
Make ERXPartialInitializer handle locking and class property settings correctly

fbarthez:
For partial entities, fix handling of the "locking" setting on attributes and the "class property" setting on attributes and relationships.

Add reminders for missing functionality to deal with user info data.


paulhoadley:
Hi Fabian,

Thanks for this. Could you tell me a little more about the problem this is solving? Was it just that those characteristics of properties weren't being merged at all?

(Not doubting your fix, it's just that we use partials quite extensively, so I keep a keen eye on changes.)

fbarthez:
Hi Paul, the problem is that on adding a new EOAttribute, it'll always be added to the list of attributes used for locking and to the list of class properties (ignoring some special cases like derived and flattened). So the fix makes sure that your choices regarding these settings are applied to the partial entity.

darkv:
@paulhoadley can we merge this?

paulhoadley:
Sorry guys, I neglected to follow this through. I'm sure we can. I'll get on it this weekend.

PR #806, opened 2016-10-02
Small patches for er.extensions.appserver.navigation

paulhoadley:
Mark Wardle supplied a couple of patches via the mailing list:

- Fixes a typo in `ERXNavigationItem.java`.
- Ensures that a generated URL in `ERXNavigationMenuItem.java` is secure if the current request is secure.

I don't use `ERXNavigationMenuItem.java`, so I can't really vouch for the latter patch, though it looks legit at a glance. Ideally someone that uses this package could check it out, though I imagine Mark's got the class in use and that this change works just fine.


wardle:
This code has been in production for two years or so and seems to work. It simply checks the current request and finds out whether it is secure returning a URL that honours the current status.

paulhoadley:
Thanks @darkv. Thanks @wardle.

PR #814, opened 2017-01-31
Add support for X-HTTP-Method-Override header on REST requests

hprange:
Sometimes a client cannot consume REST services using methods like PUT, DELETE or PATCH, due to browser limitation or firewall rules. Some frameworks provide a workaround using the X-HTTP-Method-Override header to override the method of a request. No change to the route configuration is necessary. The client of the API just informs the override header, and the ERRest library maps routes as expected. Keeping a RESTful API on the server side is the main benefit of that approach. No need for alternate / unRESTfull routes creation.

Scott Hanselman talked in more detail about that in [a blog post](http://www.hanselman.com/blog/HTTPPUTOrDELETENotAllowedUseXHTTPMethodOverrideForYourRESTServiceWithASPNETWebAPI.aspx).


hprange:
Hey guys, I'm not sure where is the best place to document this feature. Should I do that in the WOCommunity wiki?

darkv:
Hi Henrique, the WOCommunity wiki should be good for that kind of information.

PR #815, opened 2017-02-07
additional template schema for validation exceptions

darkv:
To localize an ERXValidationException there are currently five different localization keys which are checked in a specific order:

1. Entity.Property.Type
2. Entity.Property
3. Property.Type
4. Property
5. Type

The problem is if you have an entity with several properties which should give you the same localization for a specific exception type you need either:

* many similar localization entries (using schema 1.)
* change global localization for exception type (schema 5.)

The latter one avoids many duplicate keys but often you would like to restrict that localization to a specific entity and not changing it for all entities.

This patch adds a new template key _Entity.Type_ to realize that which gives a modified order:

1. Entity.Property.Type
2. Entity.Property
3. Entity.Type
4. Property.Type
5. Property
6. Type

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/validation/ERXValidationFactory.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/validation/ERXValidationFactory.java
@@ -532,6 +532,9 @@ public class ERXValidationFactory {
         // 2nd try everything minus the type.
         if (template == null)
             template = templateForKeyPath(entityName + "." + property, targetLanguage);
+        // 2.5th try entity plus type
+        if (template == null)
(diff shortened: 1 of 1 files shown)


paulhoadley:
I like it. This should be entirely backwards compatible, shouldn't it? Can you think of an edge case where this would affect an existing system? (I can't.)

darkv:
It should be safe. It is quite improbable that you get a false positive with already existing localization keys. That would be that the new _Entity.Property_ would match an already used _Entity.Type_ or _Property.Type_ geared towards a different case.

The only case I can think of is that someone uses a "manual" localization with just that exact key but then the localization probably will be the very same text.

PR #818, opened 2017-02-09
Suppress javax.naming.NameNotFoundException in servlet context 

renebock:
This patch will suppress the annoying javax.naming.NameNotFoundException when using properties to define the database connection.

Set 

   er.extensions.ERXJDBCAdaptor.className=er.extensions.jdbc.ERXJDBCAdaptor
   er.extensions.ERXJDBCAdaptor.ignoreJNDIConfiguration=true

to activate this feature.


darkv:
Hi René, do you need/want this in Wonder 6 too?

renebock:
Yes, please.

PR #819, opened 2017-02-10
Non blocking database context delegates

renebock:
When using multiple EOObjectStores, ERXMulticastingDelegate will block  potential concurrent threads.  This is a performance issue when the application is under heavy load.

This patch fix this by replacing the NSMutableArray by java.util.concurrent.CopyOnWriteArrayList and removing no longer needed synchronized blocks


renebock:
If possible, merge this patch also to wonder_6

PR #838, opened 2017-11-01
Disable the ERModernMoviesTest module temporarily from the Maven build

hprange:
I'm unable to run the integration tests on that module successfully. Neither on macOS nor AWS Linux. I'll disable the build of that module until we figure out how to make it work again.

### DIFF
--- a/Tests/pom.xml
+++ b/Tests/pom.xml
@@ -16,6 +16,7 @@
 		<module>ERXTest</module>
 		<module>TestAdaptor</module>
 		<module>PluginTest</module>
-		<module>ERModernMoviesTest</module>
+		<!-- Disable the ERModernMoviesTest module temporarily -->
+		<!--<module>ERModernMoviesTest</module>-->
 	</modules>
 </project>


hprange:
@fbarthez do ERModernMoviesTest tests pass on your machine?

fbarthez:
Hi Henrique, I haven't built wonder in quite a while. Will check it next week and let you know.

fbarthez:
Hi Henrique, I just got around to testing this. Everything works fine for me, both locally and on my jenkins server. Do you remember what failed?

hprange:
Hi, Fabian! I've spent some time investigating the cause of the problem.

**On Amazon Linux:**

The `NEXT_ROOT` environment variable wasn't set. The app was failing to start before running the tests. Setting the environment variable fixed the problem. Unfortunately, I haven't found a way to set an environment variable from the pom.xml. :|

**On my Mac:**

The app starts as expected. Even though I can access it on the browser while the tests are running, all tests fail throwing an UnreachableBrowserException. See below.

```
org.openqa.selenium.remote.UnreachableBrowserException: Could not start a new session. Possible causes are invalid address of the remote server or browser start-up failure.
```

See [ERModernMoviesTest.log](https://github.com/wocommunity/wonder/files/1455647/ERModernMoviesTest.log) for a full log of the build. Do you have any clue on what's possibly happening? Please, do not waste a lot of your time on this. Considering all the network issues brought by the last update of macOS, I'm not holding my breath.

**On Travis-CI:**

The build is working fine on Travis-CI.

**Conclusion**
All things considered, I guess we should:

1. Create a `slowTests` profile that enables the test phase of the ERModernMoviesTest module.
2. Add an enforcer rule to alert anyone trying to build this module if the NEXT_ROOT environment variable isn't set.
3. Enable the build of the ERModernMoviesTest module again.

By default, Maven won't run any tests on this project. We can enable them running Maven with `-P slowTests` argument.

What do you think? If you think that is OK, please, take a look at pull request https://github.com/wocommunity/wonder/pull/842. :)

fbarthez:
Hi Henrique! Thanks for taking the time to look into this!

The NEXT_ROOT problem's root cause (if I remember correctly) is the use of the outdated launch script in the woproject ant tasks, when no wobootstrap.jar is around at build time. There are two versions of the "appstart" script, the first one being (I think) the older one:
 * wolips/woproject-ant-tasks/src/resources/woapp/Contents/MacOS/appstart
 * wolips/woproject-ant-tasks/src/resources/woapp_52/Contents/MacOS/appstart
The second one carries on when NEXT_ROOT is not set, whereas the first one will die.

The problem on your mac seems to be that the phantomjs process dies. You could try to simply start it in the terminal and see what it complains about.

I don't really get why we would need to introduce the slowTests profile. As is, there seem to be no other integration tests around, so running maven with "verify" will cause only ERModernMoviesDemo to run in addition to the unit tests. The usual "install" or "package" should cause the unit tests to be run, but not the integration tests?

fbarthez:
I had another look and saw that a "mvn install" actually does run the integration tests. So, I'm fine with changing that behaviour as per #842.

PR #847, opened 2018-01-29
Pass MemberAccess on creation of ognl context

fbarthez:
This makes WOOgnl compatible with the 3.2 branch of ognl.


fbarthez:
This seems to cause problems, I'll have to have another look.

fbarthez:
This has since been used in production in several projects w/o problems. Looks like it was just eclipse that was acting up.

PR #852, opened 2018-02-20
Make the Ajax Modal dialog window draggable.

spelletier:


darkv:
Would it make sense to make this configurable instead of globally active?

spelletier:
I thought adding something but I did not find any use case requiring to prevent the window moving.

There is no visual changes, just the possibility to move added. Each time a new window is opened, it's position is reset so there is a safe fall back here.

darkv:
I am sure not everyone wants movable dialogs. This could even break down to make _some_ dialogs movable but not _all_. We can add this in a separate PR though.

paulhoadley:
> I am sure not everyone wants movable dialogs.

I don't use this component any more, so weight my vote accordingly, but I'd favour making this an option and preserving the old behaviour by default.

spelletier:
I added properties for default value of the movable and animation delay values.

PR #857, opened 2018-04-16
Add possibility to disable optionalPropertiesLoader method on ERXProperties class

luizalfredo23:
During a process of improving application startup time was possible to notice that the method optionalPropertiesLoader was taking a lot of time and it has not been used on the project I'm working on.

When enabled this method was called three times taking about 15 seconds each, so disabling it could save about 45 seconds at the startup.

### DIFF
--- a/Frameworks/Core/ERExtensions/Resources/Properties
+++ b/Frameworks/Core/ERExtensions/Resources/Properties
@@ -17,6 +17,11 @@ er.extensions.ERExtensions.hasLocalization=true
 ## Retain default value
 # er.extensions.ERXProperties.RetainDefaultsEnabled = true
 
+## Enables support for custom properties file lookup. 
+## E.g., Properties.[Framework], Properties.[Framework].[Username], Properties.log4j, and so on. 
+## Disabling this option might reduce the application startup time considerably.
+# er.extensions.ERXProperties.loadOptionalProperties = true
+
 ## Allows you to specify more config files (good luck:)
 # er.extensions.ERXProperties.OptionalConfigurationFiles = (/more/files)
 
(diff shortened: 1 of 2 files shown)


spelletier:
I see this method called 3 times times too but my app startup time are way shorter, usually between 5 and 10 seconds total including migration checks.

How many frameworks do you have in this app, I would expect many 100s to have some fstat take that long. Have you pinpointed the part of this method that take so long ? 15 second is an eternity for a modern computer.

I suspect this may hide another problem on your side... I ask just in case.

luizalfredo23:
Hello @spelletier, first of all, thank you for taking your time to review this pull request.
Considering what you told, I've tried to find some candidates that could cause all this startup time. 
What I found is the `optionalPropertiesLoader` method ends up reloading all the project structure many times. On another small project I've also been working on, the startup time is similar to what you mentioned above (5 to 10 sec.). But on the project I've mentioned at first, there are so many files, including ~3500 web server resources and ~500 resources, that are loaded and reloaded so many times that end up increasing startup time.

The method that reads the project structure is not on Wonder, and it's on a separate framework called `ERWebObjects`. There's no source code for this project, though.

I don't know how many developers use the optional properties on their projects. Anyway, none of my projects use them. As the method name indicates, those are optional properties. However, until now, there was no way to opt out from loading them. This pull request adds this ability.

spelletier:
I tried on a small project and my startup time went from 6 to 4 seconds. Not bad. I tried to understand how this feature is cool with no real answer. It just allows splitting a properties file in smaller subfiles. I did not knew this feature existed. I will add your new property to my defaults.

Before adding this request, I thing you should add the new property to the Properties in ERExtensions file with some comment.

BTW, if you do not uses JRebel or DCEVM with HotswapAgent you should try, it save so many app startup time by preventing them from the beginning.

spelletier:
I also included the commits to wonder 7.

PR #865, opened 2018-07-12
Jgroups multicast

NochEinKamel:
Fixes three issues we had with with jgroups multicast after migrating to wonder 7:

1. The Properties used to configure multicast Address and Port were only used in the old configuration file which was replaced during the jgroups 3.6 update. The properties had no effect anymore.
2. The Property ...localBindAddress was used to set the system property bind.address which had no effect on our system. Moved it into the configuration file as well.
3. All messages sent by one instance were received by the the same instance as well (Potentially causing an IllegalStateException - we didn't dig deeper to verify)


mkondratov:
It would be good to update @property notation and mention that jgroups.udp.mcast_port needs to be adjusted now.

PR #872, opened 2018-10-08
Adapted ExcelGenerator and ERExcelLook

maiksd:
With commit 34b653f40c9407ef3f058234ea5a92e1b74fd944, we updated Apache POI from version 3.14 to 3.17, not realizing that this included the removal of several fields that were deprecated before and now have been removed and reorganized. Those fields are accessed via KVC, so there was no compile error, and the frameworks include no tests to cover this.

This change adapts these two frameworks to use the new constants. Further testing by people that actually use these would be appreciated.

### DIFF
--- a/Frameworks/Excel/ExcelGenerator/.classpath
+++ b/Frameworks/Excel/ExcelGenerator/.classpath
@@ -7,7 +7,7 @@
 	<classpathentry exported="true" kind="con" path="WOFramework/JavaWebObjects"/>
 	<classpathentry exported="true" kind="con" path="WOFramework/JavaXML"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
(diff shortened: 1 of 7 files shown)


Frameworks/Excel/ExcelGenerator/.classpath
```diff
@@ -7,7 +7,7 @@
 	<classpathentry exported="true" kind="con" path="WOFramework/JavaWebObjects"/>
 	<classpathentry exported="true" kind="con" path="WOFramework/JavaXML"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
-	<classpathentry exported="true" kind="lib" path="Libraries/poi-3.17.jar"/>
+	<classpathentry exported="true" kind="lib" path="Libraries/poi-3.17.jar" sourcepath="/var/folders/9g/fbpw8rp13mv3bng8msrzs18w0000gn/T/.org.sf.feeling.decompiler1538989030348/source/poi-3.17-sources.jar"/>
```
darkv: That source path probably won't work for anyone but you ;-)

maiksd: Indeed :) Fixed.

PR #877, opened 2018-10-10
Add an helper method to create a ERXDatabaseAttachment from a NSData.

spelletier:


Frameworks/BusinessLogic/ERAttachment/Sources/er/attachment/processors/ERDatabaseAttachmentProcessor.java
```diff
@@ -29,6 +29,16 @@
 public class ERDatabaseAttachmentProcessor extends ERAttachmentProcessor<ERDatabaseAttachment> {
   @Override
   public ERDatabaseAttachment _process(EOEditingContext editingContext, File uploadedFile, String recommendedFileName, String mimeType, String configurationName, String ownerID, boolean pendingDelete) throws IOException {
+	  NSData data = new NSData(uploadedFile.toURI().toURL());
+	  ERDatabaseAttachment attachment = _process(editingContext, data, recommendedFileName, mimeType, configurationName);
+
+	  if (pendingDelete) {
+		  uploadedFile.delete();
```
darkv: If line 33 throws an error this line won't be reached. Prior this PR that delete would always be executed in a finally block.

PR #885, opened 2018-11-12
Fix ERXRedirect problem with component

spelletier:
Make sure appendToResponse is called before the page is accessed with the generated URL that trigger an invokeAction.

This fix problems when the page need to initialize or refresh internal state, especially for arrays in repetition. In normal usage, append to response is always called in the response before invoke action in the next request but the ERXRedirect behaviour skipped this step.

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/appserver/ERXRedirect.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/appserver/ERXRedirect.java
@@ -226,6 +226,9 @@ public class ERXRedirect extends WOComponent {
 	 */
 	public void setComponent(WOComponent component) {
 		_component = component;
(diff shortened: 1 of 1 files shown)


darkv:
This patch seems to have some annoying side effects breaking the redirection. The content sent to the client looks then like:

```
<html>
[…]
<\html>document.location.href='/Apps/WebObjects/Test.woa/wo/56.0';
```

Which prevents the change of the location as the response is mixed html and javascript and additionally sends all the html code of the current page which is not needed as the user should be immediately redirected to another page thus wasting bandwith.

In my case there is a form with a submit button on a page which calls an action that should redirect the user to another page after it's done:

```
public WOComponent save() {
  […]
  WOComponent nextPage = …;
  AjaxUtils.redirectTo(nextPage);
  return null;
}
```

PR #891, opened 2019-01-08
Revert "Fix ERXRedirect problem with component"

darkv:
This temporarily reverts the changes to ERXRedirect until problems to #885 have been sorted out.


spelletier:
I would like to investigate this but I wonder how to get raw response logged. Is there a property to enable this ? With log4j.logger.er.extensions.ERXApplication.RequestHandling=DEBUG I only get the response description (content type, length, ..) but not the actual content.

darkv:
You would have to override `dispatchRequest(WORequest)` in your application class and pipe the returned `WOResponse` to your log. I don't think that Wonder/WO has that included as that would make your log files really big in no time. Because of this the `toString()` of `WOResponse` logs only structural info.

I did check the responses from client side in the browser's web inspector where you get all info (contents, headers, …).

PR #892, opened 2019-01-15
Updated ERPDFGeneration framework to support Apache FOP 2.3.

mpugger:
- Updated ERPDFGeneration framework to support Apache FOP 2.3.
- Updated Flying Saucer to the 9.1.16 release.
- Changed from using itext to openpdf 1.2.3 (https://github.com/LibrePDF/OpenPDF/).


mpugger:
I removed the dependency because I found that the openpdf dependency was provided by the flying-saucer-pdf-openpdf dependency. 

Would it be better to restore the dependency for openpdf with a version string of 1.2.3?

darkv:
Ah, didn't see that. I will do some tests and then merge your PR. Thanks!

darkv:
Works for me.

Though if you are using TIFF images in your PDF files those won't show up anymore until you add some jars from [haraldk/TwelveMonkeys](https://github.com/haraldk/TwelveMonkeys/) to your project which restores that feature.

PR #893, opened 2019-01-21
Update prototypejs to 1.7.3

schmied:
Update prototypejs to 1.7.3


darkv:
Thanks for the PR :-)

The files `wondaculous.js` and `wondaculous-min.js` have to be updated accordingly though as those using Wonder code within Eclipse will still be using prototypejs 1.7.2 otherwise (look at Ajax/Support/compress.sh to update these files).

schmied:
Thanks for pulling! :)

PR #901, opened 2019-02-06
Fix accessing static webresources for mavenized builds

markusstoll:
As @hugith confirmed in the mailing list, the ERXStaticResourceRequestHandler does not support static resource references to resources inside a jar (as is being used with maven builds).

There is a solution with @hprange as alternative requestHandler, that you would have to register explicitly (JarResourceRequestHandler in https://gist.github.com/hprange/1068523).

This pull request extends the current ERXStaticResourceRequestHandler to be able to serve both (from WebServerResources folder or from within jar)


paulhoadley:
This PR passed me by at the time. It has taken me 19 months to realise this means I can throw out `JarResourceRequestHandler`. Nice work @markusstoll!

PR #902, opened 2019-02-06
Added missing break in switch statement for case “oracle”

rparada:
ERXQuery fix for formatting of date values for inline use with the ORACLE adaptor.

There was a break missing in the switch statement for case "oracle".  I ran into this bug when testing with ORACLE today and using inline bindings.


rparada:
This fixes the formatting of date values for inline use with the ORACLE adaptor.

rparada:
Sorry I closed by accident. This should remain open until merged into Wonder. Thanks.

PR #909, opened 2019-04-01
Apply css class for checkboxes and radio buttons

macmarco:
This code adds a css class value of "Checked" to any existing css class value. For backward-compatibility, existing css classes are not changed. Prior to this pull request, it was possible to target the selected checkbox or radio button using css, but not the table that also contained the text label.


darkv:
I think that adding a CSS class named _Checked_ is too general. I would suggest using _ERXMatrixTable--Checked_ instead—following BEM syntax—so it is clear that it is a state of the matrix table and gives you some sort of namespace as to prevent a possible name clash in existing projects.

macmarco:
Closed because we made a new pull request incorporating above comments.

PR #910, opened 2019-04-17
bugfix, indicator is only a string containing the container id

markusstoll:


markusstoll:
I dont see any errors in the travis log - and from reviewing the change I cannot see what could be a problem with it... any hints?

fbarthez:
Hi Markus, the build has been failing for a while it seems. I'm not sure about the cause, but on a hunch you might want to try and remove the .mvn/maven.config file, which configures the parallel build. That should at least make it easier to see the cause.

fbarthez:
@hprange Henrique, are you fine with disabling the parallel build feature. It looks like it's causing the build to fail since a while. If so, should we do it in a separate PR?

markusstoll:
did that, now it was building fine... strange

PR #915, opened 2019-06-13
Add generics infos to EOGlobalId conversion methods.

spelletier:


maiksd:
This does not compile. Wonder jobs in Jenkins currently fail.
https://jenkins.wocommunity.org/job/Wonder7/2549/console

PR #916, opened 2019-06-13
Add custom attribute class support in ERXQuery

spelletier:
Theses change add support of attributes using custom classes with factory and conversion methods and classes.

@rparada, can you check if these changes works with your servers. I tried on my projects with mySQL and FrontBase.


rparada:
The changes look good and I have tested them with my code without any problems.  Thanks for improving ERXQuery.

PR #917, opened 2019-06-16
Ajax interactivity improvements

markusstoll:
This is my promised pull request with the multiple update feature that I presented during WODay 2019 on April 11th in Frankfurt.

Basically I enable Ajax actions to have more than one updateContainerId as target (a comma separated string with multiple IDs or an NSArray of ids) where all referred containers will be updated in one single response.
I added an instructive example in AjaxExample app.

I understand that it will be hard to review the stuff as it goes deep into the Ajax framework. Maybe we should ask some others to take a look on it. I know @hugith  would like to use it, maybe he can take a deeper look, too.

regards, Markus


markusstoll:
About the Travis CI failure… this does not look like a problem with my code, odes it?

fbarthez:
> About the Travis CI failure… this does not look like a problem with my code, odes it?

No, that looks like a travis issue. We may have to switch to a more recent JDK. I'll try to have a closer look later today.

fbarthez:
Hi Markus, sorry for taking so long. Could you rebase your branch on the current master and then force a push? I did that and got travis to successfully build on my fork.

markusstoll:
sorry, did not work out

hugith:
Just returned from a string of summer vacations and have been putting off checking this out. Finally back at the desk and looking forward to taking a look :). Thanks for all the work Markus!

PR #918, opened 2019-06-19
Fix generics for ERXEOControlUtilities.faultsForGlobalIDs()

maiksd:


spelletier:
I suppose this fix the Jenkins compile problem. I used to compile with ant frameworks but needed to do ant all to get the error. I will adjust my process.

PR #925, opened 2020-01-02
Replace WebObjects download urls with updated ones.

spelletier:


darkv:
Hi Samuel,

can you give some explanation why those URLs should be changed? Just tested both old ones and they work as expected.

spelletier:
They did not worked when I created this pull request... I had to find out the new one.
At this time both seems to work.

darkv:
Just saw that your URLs show up on the Apple download pages. So lets use them instead :)

PR #926, opened 2020-01-20
Respecting key types for after and before comparisons

luizalfredo23:
Hello guys,

I'm trying to use the Java Time prototypes, but I've found a problem when using ERXKey after and before methods.

These methods are implemented to always compare the key type with NSTimeStamp, so when changed de key type from NSTimeStamp to LocalDateTime, for example, the implementation blocked me from that. 

When talking to @hprange about this problem he realized another problem, any type could be compared to NSTimeStamp, for instance:

```java
Person.country.after(date)
```

Even though the code above compiles, it throws an exception in runtime.

This pull request would make the two methods more flexible and ensure the compared value types are always compatible.

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/er/extensions/eof/ERXKey.java
+++ b/Frameworks/Core/ERExtensions/Sources/er/extensions/eof/ERXKey.java
@@ -1,6 +1,7 @@
 package er.extensions.eof;
 
 import java.math.BigDecimal;
+import java.time.temporal.TemporalAccessor;
 import java.util.Locale;
 
 import com.webobjects.eoaccess.EOAttribute;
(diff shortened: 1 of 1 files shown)


paulhoadley:
I like this. I take it that it's backwards compatible, in that there's no behaviour change if you _are_ supplying an `NSTimestamp`? Also, does this make these two methods just synonyms now for `greaterThan()` and `lessThan()`?

hprange:
> I take it that it's backwards compatible, in that there's no behaviour change if you are supplying an NSTimestamp?

Yes, it is as long as it has been used correctly. In other words, only with attributes of type `NSTimestamp`. Any code where the attribute type is not compatible with `NSTimestamp` will now fail at compile time. Before this change, they would fail in runtime only.

> Also, does this make these two methods just synonyms now for greaterThan() and lessThan()?

Yes, in the end, they're aliases that sound better for dates.

PR #931, opened 2020-06-13
replace blacklist and whitelist with denylist and allowlist in ERJava…

elemgee:
…Mail classes and documentation

I’ve provided cover methods for the old blacklist/whitelist methods so there can be graceful deprecation of these terms.
This isn’t a change in function at all. There has been renewed discussion in many forums about the use of blacklist and whitelist and the connotations of bad/excluded and good/included. (blacklist addresses you want to exclude and whitelist addresses are those you want to include).  Since this is a symbolic change and the use of “deny/allow” is a more clear indication of the intent and use of the functions, it’s worth making this change to these commonly used terms. Perhaps, over time, the negative associations will fade. Regardless of whether that happens, I think it’s worth using more clear terms here and being conscious of the unintended effects of some common practices.

### DIFF
--- a/Frameworks/Core/ERJavaMail/Documentation/README.txt
+++ b/Frameworks/Core/ERJavaMail/Documentation/README.txt
@@ -78,14 +78,14 @@ The X-Mailer header to put into all outgoing mail messages. Defaults to nothing.
 er.javamail.defaultEncoding = UTF-8
 The default character encoding to use for message content.  Defaults to ???.
 
-er.javamail.WhiteListEmailAddressPatterns = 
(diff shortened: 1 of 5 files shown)


rkiddy:
@elemgee So, you are not up for translating the black and white references in the japanese javadoc? :--) O well.

The changes look good and I cannot think of any reason not to merge it. Since I have not updated in a while, I will double-check and move on it soon.

Let us know if there is anything you have thought of since you set this up.

elemgee:
I’d give it a shot but I’m working mostly with Swedes these days so I can’t guarantee the translation

:-)

Larry

hprange (approved):
Looks good to me!

PR #933, opened 2020-07-04
Add convenience factory methods to create NSDictionary, NSArray, and NSSet

hprange:
As per [JEP 269](https://openjdk.java.net/jeps/269), Java 9 defined library APIs to make it convenient to create instances of collections and maps with small numbers of elements. This pull request introduces a similar concept for `NSDictionary`, `NSArray`, and `NSSet`. It provides convenience factory methods to create NSCollections in a less verbose, concise, and uniform way.

Usage:

```java
NSDictionary<String, String> dict = NSDictionary.of("k1", "v1",
                                                    "k2", "v2");

NSArray<String> array = NSArray.of("e1", "e2", "e3");

NSSet<String> set = NSSet.of("e1", "e2");
```

### DIFF
--- a/Frameworks/Core/ERExtensions/Sources/com/webobjects/foundation/NSArray.java
+++ b/Frameworks/Core/ERExtensions/Sources/com/webobjects/foundation/NSArray.java
@@ -1059,48 +1059,48 @@ public class NSArray<E> implements Cloneable, Serializable, NSCoding, NSKeyValue
 	}
 	
 	/**
-     * Returns an immutable empty {@code NSArray}.
-     *
-     * @param <E>
-     *            the {@code NSArray}'s element type
-     * @return an empty {@code NSArray}
-     */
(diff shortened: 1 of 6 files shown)


luizalfredo23:
I can't wait to save some lines of code by creating an NSArray with some items in just one line, countless times when I could have used it.

darkv:
@luizalfredo23 with the Wonder versions of those classes this was already possible since long ago like
`NSArray<String> array = new NSArray("e1", "e2", "e3");`

luizalfredo23:
@darkv 
> @luizalfredo23 with the Wonder versions of those classes this was already possible since long ago like
> `NSArray<String> array = new NSArray("e1", "e2", "e3");`

With Java 11 I've been using `List.of` so I think it would be nice using `NSArray.of`, but I did not know about the constructor, I'm using it right now. Thanks.

PR #941, opened 2021-03-01
ERAttachment name with currency error

luizalfredo23:
While uploading files with a currency symbol, an error s of `Illegal group reference` was happening.

I had to escape this symbol on each regex evaluation.


hprange (approved):

hprange:
Luiz has been using this change in production for quite some time. It seems safe to merge it.

Issue #943, opened 2021-03-23
Get Wonder building on Java 16


hugith:
Exactly what it says on the label. Not intended to actually make Wonder Java 16 compatible, just allow more recent compilers to compile it with the 1.8 source/target versions.

hugith:
Couple of things still to do:

- ERProfiling/WeavingClassLoader.java uses classes `Resource` and `URLClassPath` from `sun.misc`. These classes are available from `jdk.internal` in more recent releases. Question is: Do we want to bother or do we just move ERProfiling to "Archive"/delete it?
- ERXiss/ERXML.java has a public static variable called versionDate which uses NSTimeZone which uses internal APIs. Tests fail because of this. We _could_ create a separate maven profile to run thost tests with `--illegal-access=permit`. But do we need to? Does anyone use ERXiss or can we move that to "Archive"/delete it?

paulhoadley:
> Do we want to bother or do we just move ERProfiling to "Archive"/delete it?

As discussed elsewhere, I still use ERProfiling, so I'd vote for _bother_ if it's doable.

> Does anyone use ERXiss or can we move that to "Archive"/delete it?

My vote is for _delete._

paulhoadley:
Hey @hugith, where are we on this?

hugith:
I don't use ERPRofiling so I haven't looked into it. I opted to create my own fork instead.

paulhoadley:
Oh right, sorry—I didn't review the whole conversation above. Should we close this ticket then?

hugith:
You don’t have to keep it open on my account. Depends on if you want Wonder building on newer JDKs I guess.

paulhoadley:
We seem to have several tickets on the go that are variations on a theme. In #944 we're talking about Java 11 for `8.0-SNAPSHOT`, but I've also suggested we skip 11 and head straight to 17. Let's close this here and continue in #944.

PR #944, opened 2021-03-26
Make Wonder go to 11+6

hugithordarson:
I've been running this without problems for a couple of days now, a couple of eyes on it wouldn't hurt before it's merged.

I did not upgrade the JDK version in the "release.yml" github action to keep it compatible with 7.2. Or is that perhaps not neccessary? Not sure of the mechanics of how action configuration is handled when building older versions (does the action definiton file/jdk version come from the branch being built?)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,10 +24,10 @@ jobs:
           ${{ runner.os }}-build-
           ${{ runner.os }}-
 
-    - name: Set up JDK 11
+    - name: Set up JDK 17
       uses: actions/setup-java@v1
       with:
-        java-version: 11
+        java-version: 17
     
     - name: Deploy Snapshots with Maven
       run: mvn -B deploy --file pom.xml -s .maven_settings.xml -DretryFailedDeploymentCount=10
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,10 +24,10 @@ jobs:
           ${{ runner.os }}-build-
           ${{ runner.os }}-
 
-    - name: Set up JDK 1.8
+    - name: Set up JDK 11
       uses: actions/setup-java@v1
       with:
-        java-version: 1.8
+        java-version: 11
     
     - name: Deploy Snapshots with Maven
       run: mvn -B deploy --file pom.xml -s .maven_settings.xml -DretryFailedDeploymentCount=10
--- a/.github/workflows/manual-deploy.yml
+++ b/.github/workflows/manual-deploy.yml
@@ -18,10 +18,10 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
 
-      - name: Set up JDK 1.8
+      - name: Set up JDK 11
         uses: actions/setup-java@v1
         with:
-          java-version: 1.8
+          java-version: 11
 
       - name: Deploy snapshots with Maven
         run: mvn -B deploy --file pom.xml -s .maven-settings.xml -DretryFailedDeploymentCount=10
--- a/.github/workflows/manual-deploy.yml
+++ b/.github/workflows/manual-deploy.yml
@@ -18,10 +18,10 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
 
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v1
         with:
-          java-version: 11
+          java-version: 17
 
       - name: Deploy snapshots with Maven
         run: mvn -B deploy --file pom.xml -s .maven-settings.xml -DretryFailedDeploymentCount=10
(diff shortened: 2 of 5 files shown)


paulhoadley:
Nice work @hugith.

> a couple of eyes on it wouldn't hurt before it's merged.

As potentially the only remaining user of ERProfiling.framework, I probably need to volunteer for this. The first problem is that I don't even have Java 11 set up in _development_ yet. Two alternatives:

1. give me a little time to get set up and check out the ERProfiling.framework change; or
2. ram it through now and I'll test it later since apparently no one else uses it anyway.

What do you think?

hugith:
@paulhoadley For my parts there's no big hurry so go ahead and test :).

Before merging I'm also wondering about the consequences for the WOCommunity Jenkins builds. Jenkins probably needs to be configured with a Java 11 JDK so it can build this thing. Also; the Wonder7 job is building from master, meaning downloads for the Ant People would silently move to Java 11. We might want to have separate stable (7.2 on java 8) and master (java 11) builds there? @maiksd is probably the right person to provide input on that topic.

maiksd:
> Before merging I'm also wondering about the consequences for the WOCommunity Jenkins builds. Jenkins probably needs to be configured with a Java 11 JDK so it can build this thing. Also; the Wonder7 job there is building from master, meaning downloads for the Ant People would silently move to Java 11. We might want to have separate stable (7.2 on java 8) and master (java 11) builds there? @maiksd is probably the right person to provide input on that topic.

Jenkins already has 8 and 11 installed and configured, so that's just a matter of job configuration. But yes, the Wonder7 job uses the master branch, we should separate that before moving on. But I feel this warrants a proper discussion on the mailing list first?

hugith:
> Jenkins already has 8 and 11 installed and configured, so that's just a matter of job configuration.

Awesome!

> But yes, the Wonder7 job uses the master branch, we should separate that before moving on. But I feel this warrants a proper discussion on the mailing list first?

OK, should I start that discussion or should we first discuss what the proposal is?

hugith:
> Hi all,
> we'd like to migrate Wonder development from Java 8 to Java 11. Anyone opposed to this?

Something like this?

paulhoadley:
> Jenkins already has 8 and 11 installed and configured, so that's just a matter of job configuration. But yes, the Wonder7 job uses the master branch, we should separate that before moving on.

Here's another idea: we could start abandoning Jenkins in favour of GitHub Actions—or is the issue that we still need to support legacy Ant-based builds for the framework build products? My motivation here is just to try and reduce the number of things @maiksd needs to maintain, and Jenkins would be a nice one to ditch.

Yet another idea would be to draw an even brighter line in the sand. Wonder 8.0 would be:

* Java 11
* Maven only

Jenkins could stick around for a while to build any Wonder 7 branches, and at some point fade away. Thoughts? Too harsh?

hugith:
@paulhoadley You probably know where I stand on the Maven only idea, I think it would be fantastic and would make life easier. I believe most of us who currently work on Wonder use maven, and Ant users shouldn't be affected since the built frameworks will work fine with Ant. As for Github actions, I'm up for that too if it does everything we need. The thing I like most about Jenkins is basically @maiksd :)

maiksd:
That Jenkins is one of three that I'm maintaining. I wouldn't mind it going away, but that wouldn't actually free me from having it in my mind in general, so no big deal keeping it really. And besides Wonder, it also serves WOLips, which would need to be migrated too. And we couldn't just abandon it and then remain unable to publish minor fixes in Wonder 7 should they be needed, could we?

And I think there are still a lot of devs stuck on ant. I feel half of this community has problems moving away from what used to work for them for many years, and I'm not comfortable forcing them to.

paulhoadley:
> And besides Wonder, it also serves WOLips, which would need to be migrated too.

Yeah, I did think of WOLips as soon as I hit send on that one...

> And we couldn't just abandon it and then remain unable to publish minor fixes in Wonder 7 should they be needed, could we?

I hear what you're saying. I don't have a **strong** view on abandoning Ant, but let me push the argument a little further. Wonder isn't the vibrant open source project it used to be. There are fewer and fewer active committers, fewer mailing list participants every year. The resource that is "people who can and do work on Wonder" is stretched thinner than it ever has been. Given that picture, how long are we _obliged_ to support backwards compatibility (with JDKs, build systems)? Can we _ever_ draw a line in the sand?

These are rhetorical questions for thinking about, and aren't directed specifically at you, @maiksd!

hprange:
Moving Wonder to Java 11 would be great. After publishing the next 7.x release, we could create a new branch `wonder_7` and merge this pull request to ` master`. What do you think?

paulhoadley:
I'm all for moving ahead. Call `master` `8.0-SNAPSHOT` at that point?

paulhoadley:
Actually—hold on. At this point, why stop at 11? Should we put in the effort for 17? @hugith points out in #943 that (at least) ERProfiling.framework would be problematic. But we can't stop for a single framework.

hugith:
I think we should definitely go for 17, I only picked 11 since it was the latest LTS when I did this.

hugithordarson:
I'm going to close this PR since I believe the changes made in it were added by Ramsey in #994

PR #945, opened 2021-04-10
Nullify handling for toMany relationships.

recurve:
toMany relationships should always have an empty array. There was a bug here where we set that array to null which will surprise people. People expecting to have an empty array will have a null pointer instead without this fix.

Note: It is important to set a Mutable array because that's what it is internally even though the Apple/NeXT API shows immutable getters. If you think about it, this makes sense. They don't want you adding/remove directly from the array but internally the data structure needs to be mutable. 

Signed-off-by: Aaron Rosenzweig <aaron@chatnbike.com>


rparada:
Is this going to be merged?  I've been using it with great success.  Thanks!

rparada:
Thank you Johann

PR #956, opened 2021-06-28
ERXConfigurationManager isDeployedAsServlet relies now on the configu…

renebock:
…red WOContext class rather on a check, if the JavaWOJSPServlet framework is loaded

(Now without the ERXProperties Changes)


paulhoadley (commented):
This seems fine, but where are the new `contextClassName()` and `setContextClassName()` methods being called?

renebock:
setContextClassName is used by contextClassName for lazy initialialization, contextClassName is used by isDeployedAsServlet.

The methods may be marked as private.

paulhoadley:
Ah thanks @renebock, I was assuming there was some external usage.

paulhoadley:
I've made some trivial changes and re-submitted this work as #970, which I'll merge shortly. Thanks @renebock!

PR #958, opened 2021-10-08
Fix NullPointerException when calling the EOModel.createPrototypeCache method from the wrong EOModel (take 2)

hprange:
This error was first reported and fixed by the PR https://github.com/wocommunity/wonder/pull/551. One reason for this kind of error is the presence of duplicate `NSBundle`s in the classpath. The previous fix prevented EOModels from the same `NSBundle` from being loaded twice.

There's another scenario, though, where the same issue may arise. If two distinct `NSBundle`s have the same EOModel, the application may not initialize properly, throwing the exception below:

```
java.lang.NullPointerException
	at com.webobjects.eoaccess.EOModel.createPrototypeCache(EOModel.java:631)
	at com.webobjects.eoaccess.EOModel.prototypeAttributeNamed(EOModel.java:699)
	at com.webobjects.eoaccess.ERXModel.prototypeAttributeNamed(ERXModel.java:290)
	at com.webobjects.eoaccess.EOAttribute.<init>(EOAttribute.java:998)
```

This fix prevents EOModels with the same name from being loaded twice.


hprange:
I cherry-picked this change into the `wonder_6` branch. I've tested it with two apps: one using Wonder 7.x and the other using 6.x. No problems so far.

paulhoadley (approved):
While I haven't set up a test case which would be expected to fail without this change, I have tested this on two separate apps running Wonder 7.2. Both launch and run without issue.

PR #960, opened 2021-10-22
Fix ERJavaMail defaultSession initialization

renebock:
When er.javamail.sessionConfigViaJNDI is set to true, the application may not start in Tomcat 9. With lazy loading the defaultSession, this problem will not occur.


hprange (approved):
Looks good.

PR #962, opened 2021-11-12
Improve EOAttribute's value factory method to accept numeric arguments

hprange:
When creating custom prototypes, one can configure a factory method to convert the attribute's value from/to a string or byte array. It's not possible to use an integer or a float value as an argument/return type, though, even if you set a value type of integer ("i") or float ("f"). Trying to fetch/save an entity containing an attribute using this prototype will throw the following exception:

```
Error encountered converting value of class MyEnum to type specified in attribute 'myEnumAttribute' of entity 'MyEntity'.
Conversion exception is : EOAttribute adaptorValueByConvertingAttributeValue(Object):
Unable to convert value of class java.lang.Integer for attribute 'enumAttribute' in entity 'MyEntity' to adaptor type EOAttribute.AdaptorBytesType.
Check the signature of the conversion method MyEnum.toInteger().
```

This pull request adds the ability to configure a factory method that receives numeric values as an argument and a conversion method that returns numeric values. For instance, it allows the creation of enum prototypes that translate from/to a number in the database.

Screen Shot 2021-11-12 at 7 02 50 PM

I'm testing this change using an Oracle database, and everything has been working fine so far. I'll try to test it on other databases as well.

### DIFF
--- a/Frameworks/EOF/ERAttributeExtension/Resources/Properties
+++ b/Frameworks/EOF/ERAttributeExtension/Resources/Properties
@@ -3,4 +3,4 @@ er.extensions.load.Properties.framework.ERAttributeExtension=load
 er.extensions.ERAttributeExtension.hasLocalization=false
 
 er.extensions.ERXJDBCAdaptor.className=er.extensions.jdbc.ERXJDBCAdaptor
-er.extensions.ERXJDBCAdaptor.columnClassName=com.webobjects.jdbcadaptor.DateJDBCColumn
+er.extensions.ERXJDBCAdaptor.columnClassName=com.webobjects.jdbcadaptor.ERXEnhancedJDBCColumn
--- a/Frameworks/EOF/ERAttributeExtension/Sources/com/webobjects/eoaccess/EOAttribute.java
+++ b/Frameworks/EOF/ERAttributeExtension/Sources/com/webobjects/eoaccess/EOAttribute.java
@@ -3,10 +3,10 @@ package com.webobjects.eoaccess;
 import java.lang.reflect.InvocationTargetException;
 import java.math.BigDecimal;
 import java.sql.Timestamp;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.EnumMap;
 import java.util.Enumeration;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.TimeZone;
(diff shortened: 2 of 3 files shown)


paulhoadley (commented):
This all looks fine, though I am having trouble testing it. I have an entity with a custom `Type` enum attribute, which I've changed over to use `int4` as the external type (PostgreSQL) from the `javaEnum` prototype. I've added factory and conversion methods on `Type`:

```java
public static Type fromInt(Integer i) {
	for (Type t : values()) {
		if (t.ordinal() == i) {
			return t;
		}
	}
	return null;
}

public int toInt() {
	return ordinal();
}
```

Shouldn't that be enough to get me there? I can successfully _save_ a `Type` value to the database, but I can't read it back. Instead, I'm getting a `ClassCastExcecption` from the template method:

```
java.lang.ClassCastException: java.lang.Integer cannot be cast to net.logicsquad.juno.model.person.Employer$Type
	at net.logicsquad.juno.model.person._Employer.type(_Employer.java:187)
```

Factory method doesn't seem to get called. Any thoughts?

Screen Shot 2022-01-07 at 12 11 37 pm

hprange:
@paulhoadley It's broken. I'm already working on a fix. Thanks for taking the time to test it.

hprange:
@paulhoadley  I added the missing parts to this pull request. I tested the previous solution without ever calling the attribute getter method—shame on me. 🤦‍♂️ I'll do another round of tests this week to ensure I'm not missing something. Thanks again for taking the time to review my work. 😀

I'm still not happy with one tiny detail, though. Suppose the value factory method has a primitive type parameter (instead of a reference type). In that case, one gets an exception like the following:

```
java.lang.NoSuchMethodException: Class net.logicsquad.juno.model.person.Employer$Type does not implement method fromInt
```

This misleading error message may take lots of time for the inattentive to figure out the solution. Maybe we should log a warning before throwing the `NoSuchMethodException` just in case. Suggestions welcome.

paulhoadley:
> Suppose the value factory method has a primitive type parameter (instead of a reference type). In that case, one gets an exception like the following:

I'm running into that right now. None of `public static Type fromInt(int i)`, `public static Type fromInt(Integer i)` nor `public static Type fromInt(Number i)` seem to satisfy the search for a factory method. What's the signature it's looking for?

hprange:
@paulhoadley According to the screenshot you shared above, the value type of your property is `i` in your model. Thus, the parameter of your factory method should be of type `Integer`.

```
public static Type fromInt(Integer i)
```
I reran the tests with a nested enum type (as your example) and had no problems either.

I added a new commit to this pull request to improve the error message temporarily. Could you please try again and let me know if it works?

paulhoadley:
Thanks @hprange—works perfectly. Perhaps I had botched something myself when testing yesterday. I think this is ready to go.

paulhoadley (approved):

PR #964, opened 2021-12-04
Add a helper method to create an aggregate function for string

spelletier:
in ERXEOControlUtilities.


paulhoadley (commented):
If I'm reading this correctly, it's a cover method for `_aggregateFunctionWithQualifier()` where you supply `String.class` to that method, and then cast the result to a `String`?

hprange:
It seems so. What bothers me about this pull request is the Javadoc.

```
select MAX(AGE) from User where name like 'M*'
```
This query doesn't seem like a good example of when to use this method. @spelletier what's your use case? Can we improve the Javadoc to reflect it?

spelletier:
@paulhoadley, this is a simple wrapper indeed. @hprange, I adjusted the javadoc with a real usage.

PR #965, opened 2021-12-04
Add er.rest.format16BytesDataAsUUID property to encode 16 byte

spelletier:
NSData as pretty uuid string in json. This option only affect exactly 16 bytes data values.

Add er.rest.rfcDateFormat=rfc822 line in the default property file (with the existing defualt value) to make the existence of this property more visible.


paulhoadley (changes requested):
I don't think this should be `true` by default. I don't know how many people are emitting Base64-encoded `NSData` via REST, but anyone who is will be pretty surprised to see their 16-byte values rendered as a UUID, won't they?

spelletier:
Good remark.. Thank you. Default updated to previous behaviour.

PR #966, opened 2021-12-04
Add observeDescendentFields support in AjaxUpdateContainer 

spelletier:
to observe all descendant fields. Very useful to create updatable container that observe field using special elements like thead, tbody or tr that does not supports inner or outer div.


paulhoadley (approved):
Nice improvement. Works for me.

PR #967, opened 2021-12-04
CancellingText and uploadStarting where not using localizer like othe…

spelletier:
…r texts.

Add autocorrect = "off" in AjaxAutoComplete and AjaxDatePicker to disable browser autocorrect that fight against the components main feature...


paulhoadley (approved):
All good changes. Merge.

PR #971, opened 2022-01-16
Update log4j implementation to 2.17.1 

spelletier:
and use the slf4j bindings to replace the previous slf4j implementation.


spelletier:
I currently use this setup without problems on non D2W apps. I just see test failed, I'll check that.

hprange:
I'll check this pull request out. Thanks @spelletier. GitHub Actions builds are failing because the WOCommunity Maven repository certificate has expired. I'm working with Maik to fix it.

hprange:
@spelletier I pushed a change to your branch to fix log4j and slf4j dependencies in the pom.xml of every submodule.

renebock:
I think the hole story is a bit more complex.   I just merged your PR into my on wonder fork and I was surprised, that I didn't get a lot of compile errors.  The package org.apache.log4 is no longer available, nor does the method Logger.getLogger.

The reason is, that the JavaXML framework includes its own very ancient version of log4j1.

So, if we want get get rid of log log4j1, we have to remove all dependencies to the JavaXML.framework or provide our own version of this framework, where the old /org/apache/log4j classes are removed from the javaxml.jar

Then, we may include log4j-1.2-api-2.17.2.jar to get rid of the compile errors (see also https://logging.apache.org/log4j/2.x/manual/migration.html)

PR #981, opened 2022-05-11
Fix ERAttachment s3 url generation

spelletier:
Very old bug, probably never encountered. I found it using the S3 classes to list bucket contents and do some adjustments on headers for manually uploaded files with names in their name.


paulhoadley:
Hi Samuel,

I would like to merge this in for you. Can you walk me through the problem it's solving?

spelletier:
If the file path of the attachment contains a space, this space is incorrectly encoded by the current version and the generated url in invalid to retrieve the file.

PR #983, opened 2022-05-13
Fix error paginating queries containing the DISTINCT clause on SQL Server

hprange:
The application throws an exception when building queries with pagination for SQL Server if the query has no `ORDER BY` clause and the `DISTINCT` clause is present. This fix takes only the first column name of the query to build an `ORDER BY` clause.


darkv:
@hprange Hi Henrique, does it still make sense to cherry pick this to Wonder6?

hprange:
@darkv Nah. Don't worry about that. If someone needs it in Wonder 6, they will shout. I removed the `Wonder 6` label.

PR #984, opened 2022-05-13
Add new ERMicrosoftPlugIn supporting PK generation from sequences

hprange:
Improves the support for SQL Server with WebObjects. The `ERMicrosoftPlugIn` generates primary keys from sequences instead of the `EO_PK_TABLE` strategy used by the standard `MicrosoftPlugIn`.

Also, the original implementation of the `MicrosoftPlugIn` forcibly adds the `SelectMethod=cursor` parameter to the connection URL. Even though this option may be useful in some scenarios, it may also cause undesirable side effects. Each application may have different requirements regarding how and when to use adaptive buffering. For this reason, the `ERMicrosoftPlugin` lets the user decide how to configure adaptive buffering if needed.

See the Microsoft documentation for more information about [adaptive buffering](https://docs.microsoft.com/en-us/sql/connect/jdbc/using-adaptive-buffering).


hprange:
I've been using this plug-in for development for a while and haven't found any issues. Since this plug-in is not enabled by default, I'm willing to merge this pull request. I'm working on other improvements to this plug-in, but they deserve a separate pull request. Any objections?

PR #994, opened 2022-09-11
Mavenized wonder

nullterminated:
Migrate wonder to maven project natures.


nullterminated:
I'll leave this up a bit longer. It builds and works with JDK 18 for me.

PR #1002, opened 2022-12-21
ERXBetweenQualifier - evaluateWithObject() key path support

rparada:
ERXBetweenQualifier's evaluateWithObject() needs to use valueForKeyPath() so that it works with key paths.


paulhoadley (approved):
Looks good to me.

PR #1004, opened 2023-01-12
fix for #1004: removed restriction about bundle isJar...

sal-bock:
and uses the mainBundle if frameworkName is null.


lbane (changes requested):
This will return invalid paths for resources in JARed frameworks. This should either throw an exception or keep the current behaviour by returning null in those cases.

Issue #1005, opened 2023-01-13
ERXNavigationManager.registerObserverForFramework  throws an NullPointerException


sal-bock:
ERXNavigationManager.registerObserverForFramework  throws an NullPointerException, if the navigationMenuFile is contained in a jar.

the reason is, that ERXFileUtilities.pathForResourceNamed returns null, when the navigationMenuFile is contained in a jar.

Using pathURLForResourceNamed should fix the problem.

paulhoadley:
I assume this was fixed in #1006.

PR #1006, opened 2023-01-13
Fix for #1005: Call ERXFileUtilities.pathURLForResourceNamed instead …

sal-bock:
…of ERXFileUtilities.pathForResourceNamed.  The latter function returns null, if the resource is contained in a jar archive.


paulhoadley:
I am happy to merge this with the caveat that this is not a part of Wonder that I use, and I haven't tested the patch. If it works for @sal-bock, let's do it.

PR #1010, opened 2023-02-02
Fixed Quickstart link on README

numair:
Still not clear how a new user in 2023 starts using WO/Wonder, but it's a start... I'll try to make a guide.


numair:
Oops, I now realize a better link would be https://wiki.wocommunity.org/display/WEB/Project+Wonder+Installation -- I will re-submit with a better pull request (maybe make a few modifications to the README to make it more first-time-user friendly, and add some stuff to the homepage on the wiki).

PR #1012, opened 2023-03-24
Add WOOgnl framework

macmarco:


paulhoadley:
Hi Marco,

Could you update your pull request to include the corresponding dependency in `pom.xml`? If you don't use Maven, that's not problem—I'll test the build. (In fact, the GitHub action will test the Maven build!)

```
    <dependency>
      <groupId>wonder.core</groupId>
      <artifactId>WOOgnl</artifactId>
    </dependency>
```

macmarco:
Hi Paul, added the maven dependency as you requested. Thanks for your efforts and kind guidance!

PR #1013, opened 2023-03-25
JavaMonitor-make minor HTML/css updates to get closer to pass HTML validation

macmarco:
Add DOCTYPE to HTML; move inline styles to javamonitor.css to get closer to passing HTML validation using https://validator.w3.org/#validate_by_input


paulhoadley:
Looks good to me.

